PR #100, opened 2025-02-20
feat: [tool/googlesearch] make googleSearchRequest public

PlayerNeo42:
The [google search tool example](https://github.com/cloudwego/cloudwego.github.io/blob/e6903b254674e9dbf9456667dbedbfda3505e12d/content/zh/docs/eino/ecosystem_integration/tool/tool_googlesearch.md?plain=1#L81) on the documentation site is currently not working. Passing maps as parameters does not seem to be by design. 
Therefore, this PR renames the internal `googleSearchRequest` to `SearchRequest` to expose the interface publicly and improve its clarity.

#### What type of PR is this?

feat: make googleSearchRequest public

#### Check the PR title.

#### The PR that updates user documentation:
Examples should be updated
https://github.com/cloudwego/cloudwego.github.io/blob/e6903b254674e9dbf9456667dbedbfda3505e12d/content/zh/docs/eino/ecosystem_integration/tool/tool_googlesearch.md?plain=1#L81

### DIFF
--- a/components/tool/googlesearch/google_search.go
+++ b/components/tool/googlesearch/google_search.go
@@ -86,7 +86,7 @@ type googleSearch struct {
 	cseSvr *customsearch.Service
 }
 
-func (gs *googleSearch) search(ctx context.Context, req *googleSearchRequest) (*customsearch.Search, error) {
+func (gs *googleSearch) search(ctx context.Context, req *SearchRequest) (*customsearch.Search, error) {
 
 	num := req.Num
 	if num <= 0 {
@@ -213,7 +213,7 @@ func getDescFromPageMap(pageMap googleapi.RawMessage) (string, bool, error) {
 	return siteDesc.String(), foundDesc, nil
 }
 
-type googleSearchRequest struct {
+type SearchRequest struct {
 	Query  string `json:"query" jsonschema:"description=queried string to the search engine"`
 	Num    int    `json:"num,omitempty" jsonschema:"description=number of search results to return, valid values are between 1 and 10, inclusive"`
 	Offset int    `json:"offset,omitempty" jsonschema:"description=the index of the first result to return."`
--- a/components/tool/googlesearch/google_search_test.go
+++ b/components/tool/googlesearch/google_search_test.go
@@ -291,7 +291,7 @@ func TestGooGleSearchTool(t *testing.T) {
 
 		assert.JSONEq(t, expectedSchema, string(body))
 
-		gsr := &googleSearchRequest{
+		gsr := &SearchRequest{
 			Query: mockSearchQuery,
 			Lang:  "zh-CN",
 		}


components/tool/googlesearch/google_search.go
```diff
@@ -213,7 +213,7 @@ func getDescFromPageMap(pageMap googleapi.RawMessage) (string, bool, error) {
 	return siteDesc.String(), foundDesc, nil
 }
 
-type googleSearchRequest struct {
```
hi-pender, line 216: Sorry, I don't quite understand why we need to make googleSearchRequest public. Is this struct used outside of the package of googlesearch?

PlayerNeo42: Hi, Thanks for the reply!

While this struct may not be directly used outside the `googlesearch` package currently, I proposed this change based on the following considerations:

1. `googleSearchRequest` is actually the input parameter for the `InvokableRun()` method. While this method accepts JSON strings, lib users typically need to know the exact structure this JSON should have.

2. This approach maintains consistency with the `tool/duckduckgo` package, which I believe is the more intuitive way. As you can see in the example below, the `duckduckgo` package already exposes its SearchRequest struct publicly.

Code copied from doc site https://www.cloudwego.io/zh/docs/eino/ecosystem_integration/tool/tool_duckduckgo_search/#:~:text=//%20Create%20search%20request%0A%20%20%20%20searchReq%20%3A%3D,failed%3A%22%2C%20err)%0A%20%20%20%20%7D
```go
    // Create search request
    searchReq := &duckduckgo.SearchRequest{
       Query: "Golang programming development",
       Page:  1,
    }

    jsonReq, err := json.Marshal(searchReq)
    if err != nil {
       log.Fatal("Failed to marshal search request:", err)
    }

    // Execute search
    resp, err := tool.InvokableRun(ctx, string(jsonReq))
    if err != nil {
       log.Fatal("Search failed:", err)
    }
```

Happy to discuss further if needed!

hi-pender: I understand your perspective, but it somewhat deviates from the intended purpose of Tool.

1. the purpose of Tool is to be selected and invoked by ChatModel,  rather than serving as a utility class for developers during programming. From this perspective, exposing SearchRequest is redundant. Of course, if there is a strong need to use it as a utility class, we could consider exposing SearchRequest as a programming convention. However, at the moment, there doesn’t seem to be any indication of such a need.

2. regarding how to help developers understand the input of Tool, whether SearchRequest is exportable should have no impact.

3. DuckDuckGo exposed SearchRequest additionally for the convenience of examples and unit tests, which I believe is not a good practice. This will blur its intended purpose.

PlayerNeo42: > there doesn’t seem to be any indication of such a need

First, I want to clarify why we have a need 

1. In agent projects, we don't always rely on LLMs for everything - sometimes we do want to manually call searching, mailing, code execution, or other tools.
2. For code reuse, we tend not to re-implement google search (or any other tool) ourselves since there is already an imported package.

If we can agree on that, I believe the following will make sense

> regarding how to help developers understand the input of Tool, whether SearchRequest is exportable should have no impact.

Actually, it does impact.

Consider this:

```go
    request := map[string]any{
        "query": "Golang concurrent programming",
        "num":   3,
        "lang":  "en",
    }
    reqJSON, _ := json.Marshal(request )
    result, err := searchTool.InvokableRun(ctx, string(reqJSON))
```

I would argue this can be confusing for developers.

1. Passing a map to a function is essentially the same as passing an interface{}/any.
2. This approach basically makes a "accepting a private struct from a public function", which is generally considered bad practice.

This alternative is much cleaner:

```go
    request := &googlesearch.SearchRequest{
        Query: "Golang concurrent programming",
        Num:   3,
        Lang:  "en",
    }
    reqJSON, _ := json.Marshal(request)
    result, err := searchTool.InvokableRun(ctx, string(reqJSON))
```

With this approach, developers can not only refer to documentation but also use "Go To Definition" in their IDE to inspect the struct. More importantly, the compiler will catch typos in field names at compile-time.

> DuckDuckGo exposed SearchRequest additionally for the convenience of examples and unit tests, which I believe is not a good practice. This will blur its intended purpose.

By hiding `SearchRequest`, the assumption is made that developers will never call `InvokableRun` directly on `Tool`s. However, as I mentioned, that's not always the case.

Exposing `SearchRequest` doesn't obscure its main purpose since Eino is an agent framework. Developers familiar with the basic concepts will understand the basic scenarios for tools. If clarification is needed, it would be better to improve the documentation rather than preventing people from using it directly.

hi-pender: yes, considering that, in addition to invoking tools through LLM, there are also scenarios for manually calling the GoogleSearch tool, I agree with this AP.

hi-pender (approved):
LGTM

PR #129, opened 2025-02-27
support deepseek options

Tian2002:
#### feat:support deepseek options

In version 1.2.4 of deepseek-go, the developers added a new option called "WithPath" to set the path for the request URL. This addresses the issue where the URL of deepseek deployed on some cloud platforms differs from the official one. In the code, I added a variadic parameter list to the NewChatModel function in the deepseek package, allowing this new feature to be used in eino without any compatibility issues.


meguminnnnnnnnn:
This is a good suggestion. However, I propose that feature should be added through the addition of a config field, similar to other options like BaseURL,Model,....

PR #136, opened 2025-03-04
feat(devops): use golang code in debug input

liujian-bytedance:
#### Check the PR title.

en:
zh(optional):


shentongmartin:
Hi @Jian-Sparrow — thank you for this PR and the effort you put into it. The idea of using Go code as debug input instead of the JSON envelope format is a genuinely good improvement.

We want to be upfront: the devops module is not an active focus area for the core team right now, and we don't have the bandwidth to review and iterate on changes in this area for the foreseeable future. Rather than leave this PR open indefinitely with no realistic path to merge, we think it's more respectful of your time to close it.

If the devops module becomes a priority again in the future, we'll reference this PR as the starting point for Go code input support. If you (or anyone else) would like to maintain a fork with this feature in the meantime, that's absolutely welcome.

Thank you again for the contribution — we really appreciate it.

PR #158, opened 2025-03-14
feat: Add default base URL for Ollama

hcmlinj:
#### What type of PR is this?
According to Ollama's [API doc](https://github.com/ollama/ollama/blob/main/docs/api.md#request) we should be able to set the default base URL to `http://localhost:11434`.

#### Check the PR title.

en:
zh(optional):

### DIFF
--- a/components/model/ollama/chatmodel.go
+++ b/components/model/ollama/chatmodel.go
@@ -37,8 +37,8 @@ var CallbackMetricsExtraKey = "ollama_metrics"
 
 // ChatModelConfig stores configuration options specific to Ollama
 type ChatModelConfig struct {
-	BaseURL string        `json:"base_url"`
-	Timeout time.Duration `json:"timeout"` // request timeout for http client
+	BaseURL string        `json:"base_url"` // optional, default is http://localhost:11434
+	Timeout time.Duration `json:"timeout"`  // request timeout for http client
 
 	// HTTPClient specifies the client to send HTTP requests.
 	// If HTTPClient is set, Timeout will not be used.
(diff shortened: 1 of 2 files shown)


components/model/ollama/chatmodel.go
```diff
@@ -77,6 +77,9 @@ func NewChatModel(ctx context.Context, config *ChatModelConfig) (*ChatModel, err
 		httpClient = &http.Client{Timeout: config.Timeout}
 	}
 
+	if len(config.BaseURL) == 0 {
+		config.BaseURL = "http://localhost:11434"
```
kuhahalong, line 81: 感谢你的提交，这里我建议使用 envconfig.Host() 的方式，可见： https://github.com/ollama/ollama/blob/e27e4a3c1b9cb3a34178d5476debf805910e990a/envconfig/config.go#L19

也即如果 config 中填了 BaseURL 则使用 url.Parse, 如果没有，则使用 envconfig.Host()

components/model/ollama/chatmodel.go
```diff
@@ -37,8 +38,8 @@ var CallbackMetricsExtraKey = "ollama_metrics"
 
 // ChatModelConfig stores configuration options specific to Ollama
 type ChatModelConfig struct {
-	BaseURL string        `json:"base_url"`
-	Timeout time.Duration `json:"timeout"` // request timeout for http client
+	BaseURL string        `json:"base_url"` // optional, host can be configured via the OLLAMA_HOST environment variable otherwise default is http://localhost:11434
```
hi-pender, line 41: Ollama provides an initialization function, api.ClientFromEnvironment(), which retrieves the host from environment variables. However, as a component, it still requires initialization before use. The additional support for retrieving values from environment variables doesn’t seem to offer much convenience

hcmlinj: If you are using the default Ollama host or have already set the `OLLAMA_HOST` environment variable, there is no need to specify `BaseURL` when using `ChatModelConfig`. You only need to set `BaseURL` if you are using a non-default Ollama host and have not set the `OLLAMA_HOST` environment variable.

hi-pender: my option is that we could call api.ClientFromEnvironment() to create client when base_url is empty instead of manually reading envconfig.Host() to set base_url

PR #228, opened 2025-04-22
optimize: Improved AWS authentication when using Claude on bedrock

pintonunes:
#### What type of PR is this?

optimize

#### Check the PR title.

en:
Improvements on AWS authentication when using Claude on bedrock.
Before this change, AWS authentication could only be done using the access and the secret key passed in the model config initialization.
AWS SDK provides multiple ways to authenticate without the need to explicit set those values in the client initialization. 
With this change we can initialize the client without providing those values and the AWS SDK will perform the normal fallback authentication using environment variables, the AWS configuration file and so on.

Added the Profile config value so we can initialize the AWS client with a specific profile. Basically the same of setting the AWS_PROFILE envVar.

Make region optional. The AWS SDK can get that from envVar AWS_REGION and other places.


components/model/claude/claude.go
```diff
@@ -116,6 +121,11 @@ type Config struct {
 	// Optional for Bedrock
 	SessionToken string
 
+	// Profile is your Bedrock API AWS profile
+	// Obtain from: https://docs.aws.amazon.com/bedrock/latest/userguide/getting-started.html
+	// Optional for Bedrock
```
meguminnnnnnnnn, line 126: Please add a comment to indicate that Profile has lower priority than AKSK authentication.

pintonunes: Done

meguminnnnnnnnn (approved):
thanks for your contribution~

PR #231, opened 2025-04-26
feat: add openai reasoning effort option

mikeb26:
This parameter controls reasoning effort in the o-series models. Lower reasoning effort results in less accurate but lower latency responses, which higher reasoning effort results in the more accruate but higher latency responses. The default when not specified is currently "medium".

See
platform.openai.com/docs/api-reference/chat/create#chat-create-reasoning_effort for further detail.


libs/acl/openai/chat_model.go
```diff
@@ -321,6 +321,9 @@ func (c *Client) genRequest(in []*schema.Message, opts ...model.Option) (*openai
 		Tools:       nil,
 		ToolChoice:  c.toolChoice,
 	}, opts...)
+	openaiOpts := model.GetImplSpecificOptions(&OpenAIImplSpecificOptions{
+		ReasoningEffort: nil,
```
meguminnnnnnnnn, line 325: Do you need to use 'reasoning effort' in the call option for your scenario? I think typically using 'reasoning effort' in the config is sufficient.

mikeb26: > Do you need to use 'reasoning effort' in the call option for your scenario? I think typically using 'reasoning effort' in the config is sufficient.

Yes per-API granularity would be preferred. Having callers work around a per-config limitation seems awkward since then they would have to frequently destroy & reinit clients.

meguminnnnnnnnn: How about also adding ReasoningEffort to Config?  For users who don't need to set this in request dimensions, some simplifications can be made.

mikeb26: > How about also adding ReasoningEffort to Config? For users who don't need to set this in request dimensions, some simplifications can be made.

Added as requested & updated the PR. Sorry for the slow response.

libs/acl/openai/chat_model.go
```diff
@@ -145,6 +145,10 @@ type Config struct {
 	// ExtraFields will override any existing fields with the same key.
 	// Optional. Useful for experimental features not yet officially supported.
 	ExtraFields map[string]any `json:"-"`
+
+	// ReasoningEffort will override the default reasoning level of "medium"
+	// Optional. Useful for fine tuning response latency vs. accuracy
+	ReasoningEffort *ReasoningEffortLevel
```
meguminnnnnnnnn: This field doesn't need to be a pointer because the null value and the default value are equivalent.

meguminnnnnnnnn (approved):

PR #239, opened 2025-04-30
fix: Claude tool call when the tool has no arguments

### LINKED ISSUE #238: Claude: Tools without arguments
When using Claude with Tools that dont have arguments, the eino tool component fails with "failed to unmarshal arguments in json" (empty string)

This happens cause the empty json string is stripped from the arguments. 
https://github.com/cloudwego/eino-ext/blame/a2545a66f5cf63a9a582182fbff72bf752ec38c9/components/model/claude/claude.go#L625

This only happens when streaming and I understand why you removing the `{}`.

I dont really know where to fix this. But probably add here:
https://github.com/cloudwego/eino/blob/e17d69d5a166100226cd1213c9069879438a55d5/components/tool/utils/invokable_func.go#L135

a check like, if arguments == "" then arguments == "{}"

pintonunes:
#### Check the PR title.

en:
- When the tool to be called has no arguments, the message outputted by the ChatModel has an empty string in the function arguments which makes the Tool component to fail.

Example:

```
=====================================================================================================
[OnStart] Component: ToolsNode, Name: Tools, Type: 
 {
  "role": "assistant",
  "content": "I'll help you get information .......",
  "tool_calls": [
    {
      "index": 0,
      "id": "toolu_bdrk_01JTRPBZ46wZqEnfBurtM9sJ",
      "type": "",
      "function": {
        "name": "get_current_time_rfc3339"
      }
    }
  ],
  "response_meta": {
    "finish_reason": "tool_use",
    "usage": {
      "prompt_tokens": 5258,
      "completion_tokens": 70,
      "total_tokens": 5260
    }
  }
}
=====================================================================================================
[OnStart] Component: Tool, Name: get_current_time_rfc3339, Type: GetCurrentTimeRfc3339
 ""
=========[OnError]=========
[LocalFunc] failed to unmarshal arguments in json, toolName=get_current_time_rfc3339, err="Syntax error no sources available, the input json is empty: errors.SyntaxError{Pos:0, Src:\"\", Code:0x1, Msg:\"\"}"
```
   

Fixes #238

### DIFF
--- a/components/model/claude/claude.go
+++ b/components/model/claude/claude.go
@@ -254,12 +254,17 @@ func (cm *ChatModel) Stream(ctx context.Context, input []*schema.Message, opts .
 				waitList = append(waitList, message)
 				continue
 			}
+			if len(message.ToolCalls) > 0 {
+				waitList = append(waitList, message)
+				continue
+			}
 			if len(waitList) != 0 {
 				message, err = schema.ConcatMessages(append(waitList, message))
 				if err != nil {
 					_ = sw.Send(nil, fmt.Errorf("concat empty message fail: %w", err))
 					return
 				}
+				fixToolsWithEmptyArgs(message)
 				waitList = []*schema.Message{}
 			}
 			closed := sw.Send(cm.getCallbackOutput(message), nil)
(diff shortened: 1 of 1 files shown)


meguminnnnnnnnn:
This handling will eliminate the streaming effect of tool call arguments. 
I came up with a solution: we can record the arguments of each tool call at the beginning, and if we know that the arguments are all empty at the end of the tool call (or at the start of the next tool call), we can manually add a frame with arguments {} for this tool call. Would this be feasible?

yclw:
Should we also add fixToolsWithEmptyArgs(message) before callbacks.OnEnd() in the Generate method? I'm still getting "unexpected end of JSON input" errors with non-streaming calls using the current fix.

mrh997:
@pintonunes The `Generate` method also needs to be fixed.

shentongmartin:
This is not a concern for ChatModel, but Tool. The recommended way to solve this is to use too middleware.

PR #241, opened 2025-05-01
Fix: incorrect dependency version for go-openai in embedding/openai

Wood-Q:
This PR fixes an incorrect dependency version in `components/embedding/openai/go.mod`.

The original version of `github.com/meguminnnnnnnnn/go-openai` was:

```
v0.0.0-20250402131905-e1ff67830216
```

However, this revision `e1ff67830216` does not exist, causing the following error when attempting to build or import the package:

```
invalid version: unknown revision e1ff67830216
```

I have updated the version to a valid commit:

```
v0.0.0-20250408071642-761325becfd6
```

This resolves the module resolution error and allows the project to build and test properly.

---

**Fixes:** #240

---


Wood-Q:
I mistakenly submitted two PRs as one commit.

PR #295, opened 2025-06-05
fix: mcp InvokableRun int64 precision loss

FH-30:
#### What type of PR is this?

fix: mcp InvokableRun int64 precision loss

#### Check the PR title.

#### More detailed description for this PR:
Sonic by default unmarshals a number encountered in a json string into type float64 causing precision loss when the number needs more than 53 bits to represent it, this PR enables the UseNumber config of sonic to unmarshal any encountered number in the json string into a json.Number instead which preserves both int64 and float64 preventing any precision loss.


components/tool/mcp/mcp.go
```diff
@@ -96,7 +96,7 @@ func (m *toolHelper) Info(ctx context.Context) (*schema.ToolInfo, error) {
 
 func (m *toolHelper) InvokableRun(ctx context.Context, argumentsInJSON string, opts ...tool.Option) (string, error) {
 	arg := make(map[string]any)
-	err := sonic.Unmarshal([]byte(argumentsInJSON), &arg)
+	err := sonic.Config{UseNumber: true}.Froze().Unmarshal([]byte(argumentsInJSON), &arg)
```
meguminnnnnnnnn: The type of CallToolRequest.Params.Arguments is map[string]any in early version, so i unmarshal argumentsInJson to map first. 
And in the latest version, the type has been modified to any. So you can pass argumentsInJson directly without unmarshal.

FH-30: @meguminnnnnnnnn if you pass in the json string directly into any without unmarshalling into a map first this will break

top rectangle is after marshalling into a map[string]any first, the second rectangle is directly passing it as a string skipping the unmarshal, notice they do not produce the same call tool request sent (this is logged out on python mcp server side)
image

this is because when the go mcp client sdk marshals the call tool request to send to the server, marshalling a go map in CallToolRequest.Params.Arguments will preserve it being marshalled as a json map and the arguments of the call tool must be a json map. Passing in the go string directly into the json marshal when marshalling CallToolRequest.Params.Arguments will cause it to be marshalled as a json string breaking it, I tested this by modifying the eino-ext as follows, it breaks

image

meguminnnnnnnnn: Converting it to json.RawMessage can resolve this problem

FH-30: @meguminnnnnnnnn this solves it, thanks

meguminnnnnnnnn:
> [CLA assistant check](https://cla-assistant.io/cloudwego/eino-ext?pullRequest=295) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/cloudwego/eino-ext?pullRequest=295) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/cloudwego/eino-ext?pullRequest=295) it.

需要签下 cla

meguminnnnnnnnn (approved):

Issue #320, opened 2025-06-21
Xslx Parser - Content<=>Meta config


eugeis:
It would be great to have configurable column handling, specifying which columns are joined to content, stored in metadata, or ignored.  The default behavior should remain joining and storing all columns as content unless a configuration is provided.  The configuration structure could be:

```
type Columns struct {
    Content []string // e.g., ["A", "D", "F"]
    Meta    []string // e.g., ["B", "C"]
}
```

Thank You

tangchaojun-bytedance:
thank you for your advice.

PR #393, opened 2025-08-06
feat: Enhanced XLSX parser with runtime configuration and custom columns processing

eugeis:
#### What type of PR is this?
feat: Enhanced XLSX parser with runtime configuration and custom columns processing

#### Check the PR title.
- [x] This PR title match the format: <type>(optional scope): <description>
- [x] Attach the PR updating the user documentation if the current PR requires user awareness at the usage level.

en:
This PR enhances the XLSX parser component with the following features:

1. **Runtime Configuration**: Allows overriding the default parser configuration at runtime using the `WithConfig` option
2. **Custom Columns Processing**:
   - Selectively use specific columns for document content via `Columns.Content`
   - Define which columns should be included in metadata via `Columns.Meta`
   - Rename columns with custom names via `Columns.CustomNames`
   - Full support for combining these options for flexible data extraction
   
   
type Columns struct {
	// NoHeader is set to false by default, which means that the first row is used as the table header
	NoHeader bool

	Content     []string          // e.g., ["A", "D", "F"]
	Meta        []string          // e.g., ["B", "C"]
	CustomNames map[string]string // e.g., {"A": "Name", "B": "Age"}
}

These enhancements enable more precise control over how data is extracted from XLSX files, making the parser more versatile for different document processing needs.

### DIFF
--- a/components/document/parser/xlsx/xlsx_parser.go
+++ b/components/document/parser/xlsx/xlsx_parser.go
@@ -42,21 +42,21 @@ type XlsxParser struct {
 
 type Columns struct {
 	// NoHeader is set to false by default, which means that the first row is used as the table header
-	NoHeader bool
+	NoHeader bool `yaml:"no_header" json:"no_header"`
 
-	Content     []string          // e.g., ["A", "D", "F"]
-	Meta        []string          // e.g., ["B", "C"]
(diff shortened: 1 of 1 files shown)


eugeis:
See #320

PR #399, opened 2025-08-08
feat(claude): require at least 1 user msg

shentongmartin:
#### What type of PR is this?
feat: add msg checks for claude
1. require at least 1 user msg
2. user msg should be the first msg other than system
3. can have more than one system msg


components/model/claude/claude.go
```diff
@@ -395,11 +395,40 @@ func toAnthropicToolParam(tools []*schema.ToolInfo) ([]anthropic.ToolUnionParam,
 	return result, nil
 }
 
-func (cm *ChatModel) genMessageNewParams(input []*schema.Message, opts ...model.Option) (anthropic.MessageNewParams, error) {
+func preProcessMessages(input []*schema.Message) ([]*schema.Message, []*schema.Message, error) {
+	userMsgIdx := -1
+	for i, msg := range input {
```
meguminnnnnnnnn, line 400: user、system、user 这种也可以吗

shentongmartin: 这里面没校验

PR #449, opened 2025-09-15
fix(embedding): Add batch size configuration to support batch

biaode:
为openai的embedding组件增加一个可选的BatchSize参数
作用：避免传入长文本列表时，API服务端返回超出并发处理数量限制问题。

### DIFF
--- a/components/embedding/openai/embedding.go
+++ b/components/embedding/openai/embedding.go
@@ -85,6 +85,10 @@ type EmbeddingConfig struct {
 	// User is a unique identifier representing your end-user
 	// Optional. Helps OpenAI monitor and detect abuse
 	User *string `json:"user,omitempty"`
+
+	// BatchSize specifies the number of texts to embed in a single request
+	// Optional.
+	BatchSize int `json:"batch_size,omitempty"`
 }
 
 var _ embedding.Embedder = (*Embedder)(nil)
@@ -114,6 +118,7 @@ func NewEmbedder(ctx context.Context, config *EmbeddingConfig) (*Embedder, error
 			EncodingFormat: config.EncodingFormat,
 			Dimensions:     config.Dimensions,
 			User:           config.User,
+			BatchSize:      config.BatchSize,
 		}
 	}
 	cli, err := openai.NewEmbeddingClient(ctx, nConf)
--- a/libs/acl/openai/embedding.go
+++ b/libs/acl/openai/embedding.go
@@ -20,10 +20,9 @@ import (
 	"context"
 	"net/http"
 
-	"github.com/meguminnnnnnnnn/go-openai"
-
 	"github.com/cloudwego/eino/callbacks"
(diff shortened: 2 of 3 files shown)


shentongmartin:
Thank you for the contribution! After careful review, we believe this change is better handled at the caller/application level rather than inside the framework component. Here is our reasoning:

## 1. Callback & Observability Semantics Break

The current implementation fires `callbacks.OnStart` once with all texts, but then makes multiple API calls internally. If a failure occurs mid-batch, the callback system cannot report partial completion. Each batch should ideally be a separate observable unit — with its own span in traces, its own latency measurement, and its own error reporting.

## 2. The Codebase Already Follows the Correct Pattern

Existing indexers (Qdrant, Redis, VikingDB) already handle batching at the **caller level** — they iterate over documents in batches and call `emb.EmbedStrings()` per batch. This is the right approach because the caller has business context: it knows what error semantics it wants (partial success? all-or-nothing?), and it can coordinate batching across multiple operations (embed + store).

## 3. Different From Provider-Level API Constraints

The existing batching in TencentCloud's embedder (`batchSize := 200`) works around a **hard API constraint** (Hunyuan rejects >200 inputs). That's an invisible provider limitation. A user-configurable `BatchSize` is a **policy decision** — a fundamentally different concern that belongs in user code, not the component.

## 4. Trivially Achieved in User-Land

Users can achieve this in ~10 lines with **full control** over error handling, concurrency, and retry:

```go
func embedInBatches(ctx context.Context, emb embedding.Embedder, texts []string, batchSize int) ([][]float64, error) {
    result := make([][]float64, 0, len(texts))
    for i := 0; i < len(texts); i += batchSize {
        end := min(i+batchSize, len(texts))
        vectors, err := emb.EmbedStrings(ctx, texts[i:end])
        if err != nil {
            return result, fmt.Errorf("batch %d-%d failed: %w", i, end, err)
        }
        result = append(result, vectors...)
    }
    return result, nil
}
```

Or as a decorator (like the existing `cache.Embedder` pattern):

```go
type BatchEmbedder struct {
    inner     embedding.Embedder
    batchSize int
}

func (b *BatchEmbedder) EmbedStrings(ctx context.Context, texts []string, opts ...embedding.Option) ([][]float64, error) {
    // split and delegate to b.inner.EmbedStrings per batch
}
```

Both approaches give correct callback semantics (one OnStart/OnEnd per batch), work with the cache wrapper, and let users control concurrency and error policy.

## 5. Maintenance Burden Across Providers

If added to OpenAI, the same `BatchSize` logic would need to be replicated in Ark, Gemini, Ollama, DashScope, Qianfan, etc. A decorator or caller-side utility solves it once for all providers.

---

We appreciate the intent behind this PR. If this is a common need, the better path would be a shared `BatchEmbedder` decorator (similar to `cache.Embedder`) that wraps any `embedding.Embedder` — we'd welcome a PR for that approach!

PR #482, opened 2025-10-16
fix(model_claude): tool call with empty arguments

### LINKED ISSUE #238: Claude: Tools without arguments

mrh997:
#### Check the PR title.

en:
zh(optional): 

Fixes #238 #462

### DIFF
--- a/components/model/claude/claude.go
+++ b/components/model/claude/claude.go
@@ -219,15 +219,19 @@ func (cm *ChatModel) Generate(ctx context.Context, input []*schema.Message, opts
 	if err != nil {
 		return nil, err
 	}
+
 	resp, err := cm.cli.Messages.New(ctx, msgParam)
 	if err != nil {
 		return nil, fmt.Errorf("create new message fail: %w", err)
 	}
+
 	message, err = convOutputMessage(resp)
 	if err != nil {
 		return nil, fmt.Errorf("convert response to schema message fail: %w", err)
 	}
+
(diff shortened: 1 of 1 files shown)


hi-pender (approved):

components/model/claude/claude.go
```diff
@@ -685,9 +691,15 @@ func convSchemaMessage(message *schema.Message) (mp anthropic.MessageParam, err
 	}
 
 	for i := range message.ToolCalls {
-		messageParams = append(messageParams, anthropic.NewToolUseBlock(message.ToolCalls[i].ID,
-			json.RawMessage(message.ToolCalls[i].Function.Arguments),
-			message.ToolCalls[i].Function.Name))
+		tc := message.ToolCalls[i]
```
meguminnnnnnnnn, line 694: 这个也加个主食吧

mrh997: done

PR #498, opened 2025-10-23
fix(model_arkbot): registered the same type as arkbot

mrh997:
#### Check the PR title.

en: This is a breaking change. Because ArkBot registers the same type name as ARK, a panic will occur if both ArkBot and the ARK module are referenced at the same time.
zh(optional):

### DIFF
--- a/components/model/arkbot/message_extra.go
+++ b/components/model/arkbot/message_extra.go
@@ -24,12 +24,19 @@ import (
 )
 
 const (
-	keyOfRequestID           = "ark-request-id"
+	// Deprecated: use keyOfBotRequestID instead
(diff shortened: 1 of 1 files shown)


components/model/arkbot/message_extra.go
```diff
@@ -61,82 +68,73 @@ func init() {
 		return ret, nil
 	})
 
-	schema.RegisterName[arkRequestID]("_eino_ext_ark_request_id")
+	schema.RegisterName[arkRequestID]("_eino_ext_ark_bot_request_id")
```
shentongmartin, line 71: 这是得注册 keyOfBotRequestID?

shentongmartin (approved):

PR #503, opened 2025-10-27
Fix/sandbox avoid duplicate workspace mount

dipfocus:
- The sandbox could create a temporary host->container bind for the workdir (default "/workspace").
- This could occur even when the caller already provided a host->container binding to the same container path.
- The double-bind caused Docker to fail with "Duplicate mount point: /workspace".
- This PR updates prepareVolumeBindings to skip the auto-created workdir bind when a user-provided binding already targets the configured WorkDir.

### DIFF
--- a/components/tool/commandline/sandbox/sandbox_test.go
+++ b/components/tool/commandline/sandbox/sandbox_test.go
@@ -97,6 +97,22 @@ func TestDockerSandbox_Create(t *testing.T) {
 	assert.Equal(t, "test_container_id", sandbox.containerID)
 }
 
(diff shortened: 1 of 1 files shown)


meguminnnnnnnnn (approved):
LGTM

meguminnnnnnnnn:
Thanks for the PR! The changes look good overall.
Could you please add or update the unit tests to ensure the new logic is properly covered?
This will help us maintain consistent test coverage before merging.

dipfocus:
Added prepareVolumeBindings regression test to cover the new skip-auto-mount path.

PR #512, opened 2025-10-30
feat(gemini): Use default value for GEMINI_MODEL in example/generate

dipfocus:
When the GEMINI_MODEL environment variable is not set in the generate example, it now defaults to gemini-2.5-flash. This prevents errors when running the example without this environment variable configured.

### DIFF
--- a/components/model/gemini/examples/generate/generate.go
+++ b/components/model/gemini/examples/generate/generate.go
@@ -30,7 +30,10 @@ import (
 
 func main() {
 	apiKey := os.Getenv("GEMINI_API_KEY")
-	modelName := os.Getenv("GEMINI_MODEL")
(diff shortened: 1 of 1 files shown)


components/model/gemini/examples/generate/generate.go
```diff
@@ -30,7 +30,10 @@ import (
 
 func main() {
 	apiKey := os.Getenv("GEMINI_API_KEY")
-	modelName := os.Getenv("GEMINI_MODEL")
+	modelName, ok := os.LookupEnv("GEMINI_MODEL")
+	if !ok {
```
JonXSnow, line 34: exmples 暂时不需要这个判断，只是一个示例，都保持风格统一，如果modelname 为空模型侧报错兜底吧

PR #621, opened 2025-12-22
修复：确保在配置中未指定 HTTPClient 时使用默认客户端

CengSin:
#### Check the PR title.

en:
zh(optional):


shentongmartin:
Closing this PR — the original code is actually correct.

The `config` parameter is a pointer passed in by the caller and is **not owned by `NewEmbeddingClient`**. Mutating `config.HTTPClient` is a side effect that the caller does not expect and could break downstream logic (e.g., if the caller checks `config.HTTPClient == nil` to determine whether a custom client was provided).

The original pattern:

```go
clientConf.HTTPClient = config.HTTPClient
if clientConf.HTTPClient == nil {
    clientConf.HTTPClient = http.DefaultClient
}
```

correctly applies the default to the **local** `clientConf` only, without modifying the caller's input. This is the right approach.

PR #624, opened 2025-12-24
feat(retriever): 添加返回字段和文档转换器配置支持

CengSin:
Add support for custom queries of the payload attribute in qdrant


shentongmartin:
## Thorough Review of PR #624: Qdrant Retriever — Return Fields & Document Converter Support

### Value Proposition

This PR addresses a **real gap** in the Qdrant retriever's API surface. After examining all 10+ retriever implementations in eino-ext, the Qdrant retriever was the **only major retriever** that lacked a configurable document converter:

| Retriever | Has configurable converter? | Field name |
|---|---|---|
| ES7, ES8, ES9, OpenSearch2 | ✅ | `ResultParser` |
| Milvus v1, Milvus v2 | ✅ | `DocumentConverter` |
| Redis | ✅ | `DocumentConverter` + `ReturnFields` |
| **Qdrant (before PR)** | ❌ | N/A — hardcoded inline |

The `DocumentConverter` addition is clearly valuable and the naming is consistent with Milvus/Redis. The improved ID handling (supporting both UUID and numeric IDs) is also a real bug fix.

---

### 🔴 Critical Issues

**1. `ReturnFields` is NOT passed to Qdrant at the protocol level**

The `ReturnFields` config is only used at the *parsing* level (in `defaultResultParser`), but the actual Qdrant query still fetches **all** payload fields:

```go
// Still unchanged in Retrieve():
searchReq := qdrant.QueryPoints{
    WithPayload: qdrant.NewWithPayload(true), // ← Fetches ALL payload fields
}
```

Compare with the Redis retriever, which correctly propagates `ReturnFields` to the `FT.SEARCH RETURN` clause. The Qdrant Go client supports `PayloadIncludeSelector` via `WithPayloadSelector`. The fix should be:

```go
WithPayload: &qdrant.WithPayloadSelector{
    SelectorOptions: &qdrant.WithPayloadSelector_Include{
        Include: &qdrant.PayloadIncludeSelector{Fields: returnFields},
    },
}
```

Without this, `ReturnFields` is misleading — the name implies a network-level optimization but it's purely cosmetic filtering that still fetches all data over the wire.

**2. `defaultResultParser` hard-fails on missing fields — behavioral regression**

```go
val, found := point.Payload[field]
if !found {
    return nil, fmt.Errorf("[defaultResultParser] field=%s not found in payload, point=%v", field, point)
}
```

The default `ReturnFields` is `["metadata", "content"]`. Many users only store `"content"` and other flat fields without a `"metadata"` key. The old code handled this gracefully with `if val, ok := pt.Payload[key]; ok` (missing fields silently skipped). This new behavior will **break existing users**.

Note: The existing `TestRetrieverRetrieve` test mock only contains `defaultContentKey` (no `defaultMetadataKey`), which demonstrates this regression — the test would fail with the new code.

**3. Config mutation — modifying the caller's config object**

```go
if len(config.ReturnFields) == 0 {
    config.ReturnFields = []string{defaultMetadataKey, defaultContentKey}
}
```

This mutates the `*Config` passed by the caller. If a caller reuses their config (e.g., creating multiple retrievers with variations), subsequent calls see already-modified values. The fix: copy locally, don't modify `config`.

---

### 🟡 Design Concerns

1. **`ReturnFields` comment is confusing**: `"num is the number of attributes following the keyword"` reads like Redis `RETURN <num> <field1>` syntax documentation — not applicable to a Go `[]string`.

2. **Default order mismatch**: Doc says `Default []string{"content", "metadata"}` but code is `[]string{defaultMetadataKey, defaultContentKey}` (metadata first).

3. **`ReturnFields` silently ignored with custom `DocumentConverter`**: When a user provides their own converter, `ReturnFields` is completely dead code. Neither doc nor code communicates this coupling.

4. **`returnFields` not stored on `Retriever` struct**: Only `documentConverter` is stored. This prevents using `ReturnFields` to optimize the Qdrant query (Issue 1) and makes the retriever opaque to inspection.

5. **Missing `default` case in type switch**: If `val.GetKind()` returns `nil` (possible with protobuf zero values), the field is silently dropped. A `default` case should handle this.

---

### 📝 Missing Tests

The PR modifies behavioral code but includes no test updates:
- Existing `TestRetrieverRetrieve` would break (mock lacks `"metadata"` key)
- No tests for custom `ReturnFields`
- No tests for custom `DocumentConverter`
- No tests for numeric ID handling

---

### ✅ What's Done Well

- `defaultResultParser` as a closure factory (matches Redis pattern)
- Comprehensive protobuf type handling for all `qdrant.Value` variants
- Nil point/document guards
- Score applied outside converter (correct separation of concerns)

---

### Recommendations

1. Use `WithPayloadSelector` + `PayloadIncludeSelector` to limit what Qdrant sends over the wire
2. Make missing fields soft-fail (skip, don't error) matching original behavior
3. Don't mutate `config` — copy `ReturnFields` locally
4. Fix the doc comment (remove Redis-specific language)
5. Store `returnFields` on the `Retriever` struct
6. Add a `default` case to the type switch
7. Add comprehensive tests

**Verdict**: The `DocumentConverter` addition is clearly valuable and brings Qdrant in line with other retrievers. However, the `ReturnFields` implementation has gaps (not propagated to protocol), the hard-fail on missing fields is a regression, and tests are needed. I'd recommend addressing the critical issues before merging.

tangchaojun-bytedance:
@CengSin 单测报错，请解决。

CengSin:
@tangchaojun-bytedance 错误已经修复

tangchaojun-bytedance:
@CengSin  有以下问题，请你思考，修改或解释：
1、修改后Qdrant 仍把所有 payload 发回来，ReturnFields 只是让 defaultResultParser 从 point.Payload 里读哪些字段。这不是真正的“返回字段限制”；Qdrant Go SDK 里 NewWithPayload(true) 是 “enable payload”，也就是返回全部 payload；字段 include 要用 NewWithPayloadInclude(...)，应该直接用qdrant的返回字段限制的接口；
2、修改前 main 对 content / metadata 都是存在才解析；修改后 默认 ReturnFields={metadata,content}，任何一个不存在都会 error，是否合理？
3、自定义 DocumentConverter 时 ReturnFields 实际无效，因为请求仍是 NewWithPayload(true)，和问题1类似；
4、请顺带修改对应的readme文件的描述。

tangchaojun-bytedance:
@CengSin 感谢您迅速的处理，这里有些小问题。
1、英文readme文件中DocumentConverter 示例有误示例读取 point.Payload["text"]，但没有配置 ReturnFields: []string{"text"}；当前实现默认只请求 metadata/content，用户照抄会拿不到 text；
2、NewRetriever 仍会修改调用方传入的 config.DocumentConverter
retriever.go (line 86) 这里会把默认 converter 写回 config。若调用方复用同一个 config 并改 ReturnFields 再创建第二个 retriever，converter 仍可能闭包旧字段，和新的 returnFields 不一致。建议改成本地变量。
3、中文的readme文件也要修改。

CengSin:
> @CengSin 感谢您迅速的处理，这里有些小问题。 1、英文readme文件中DocumentConverter 示例有误示例读取 point.Payload["text"]，但没有配置 ReturnFields: []string{"text"}；当前实现默认只请求 metadata/content，用户照抄会拿不到 text； 2、NewRetriever 仍会修改调用方传入的 config.DocumentConverter retriever.go (line 86) 这里会把默认 converter 写回 config。若调用方复用同一个 config 并改 ReturnFields 再创建第二个 retriever，converter 仍可能闭包旧字段，和新的 returnFields 不一致。建议改成本地变量。 3、中文的readme文件也要修改。

已修改，我没考虑到对config的修改会影响config复用的情况，感谢您的review。

PR #635, opened 2026-01-02
feat(claude): add MaxRetries configuration option

pintonunes:
## Summary

Add support for configuring retry behavior through the MaxRetries option in all Claude model implementations (Direct API, Vertex AI, and AWS Bedrock).

## Changes

- Added `MaxRetries *int` field to the `Config` struct
- Updated `NewChatModel` to apply `option.WithMaxRetries()` for all three implementations:
  - Direct Anthropic API
  - Google Vertex AI  
  - AWS Bedrock
- Added comprehensive documentation explaining:
  - Default behavior: 2 retries if not specified
  - Set to 0 to disable retries
  - Set to positive integer for custom retry count

## Testing

- All existing tests pass
- Build successful for all implementations

## Usage Example

```go
maxRetries := 3
config := &claude.Config{
    APIKey:     "your-api-key",
    Model:      "claude-3-5-sonnet-20241022",
    MaxTokens:  2000,
    MaxRetries: &maxRetries, // Configure retry behavior
}
model, _ := claude.NewChatModel(ctx, config)
```


shentongmartin:
Thank you for the contribution! However, after careful analysis, we've decided not to merge this PR. Here's the reasoning:

**The Anthropic SDK's `MaxRetries` only retries pre-stream HTTP errors, not mid-stream errors.**

The Anthropic Go SDK's retry mechanism (via `option.WithMaxRetries`) operates at the HTTP transport level. It retries on connection errors, HTTP 408, 409, 429, and >=500 status codes. However, once the server returns HTTP 200 and SSE streaming begins, the SDK commits to that stream — any error that occurs mid-stream (e.g., an SSE `error` event, a TCP connection drop, or malformed JSON in a chunk) is simply surfaced as `stream.Err()` with **no retry**.

**Eino's ADK already provides strictly superior retry coverage.**

The eino ADK's `ModelRetryConfig` (in `adk/retry_chatmodel.go`) implements a full stream-aware retry via a buffered-consume-then-replay strategy:

1. It forks the stream into two copies
2. Eagerly consumes one copy to check for errors in every chunk
3. If the entire stream succeeds → returns the buffered copy
4. If any chunk has an error → discards the stream and retries the entire request

This covers **all** failure scenarios — both pre-stream errors (which the Anthropic SDK can handle) and mid-stream errors (which it cannot).

**Adding SDK-level `MaxRetries` would be redundant and potentially harmful:**

- For pre-stream errors: eino's `retryModelWrapper` already retries them, making SDK-level retry redundant.
- For streaming errors: the SDK's retry can't help here anyway — only eino's retry can.
- With both enabled, a transient 429 would trigger N SDK retries × M eino retries = N×M total attempts with compounded delays, which is likely not the intended behavior.

For users who need retry, configuring `ModelRetryConfig` at the eino ADK level is the correct and complete approach.

PR #688, opened 2026-02-10
fix(langfuse): fix latency display showing incorrect huge hours value

favbox:
## Problem
Langfuse UI was displaying incorrect latency values like **"17751749h 37m 57s"** 
instead of the actual duration in milliseconds (e.g., "854ms").

## Root Cause
The callback handler was not preserving the original `startTime` when 
updating spans/generations in `OnEnd`. This caused the `startTime` to be 
reset to zero value (`"0001-01-01T00:00:00Z"`), which Langfuse backend 
uses to calculate latency as `endTime - startTime`, resulting in a 
massive time difference.

## Solution
1. Add `Duration` field (milliseconds) to `SpanEventBody` and 
   `GenerationEventBody` in event data model
2. Store `startTime` in callback state during `OnStart`
3. Preserve original `startTime` and calculate `duration` in `OnEnd`

## Changes
- `libs/acl/langfuse/event.go`: Add `Duration *int64` field (milliseconds)
- `callbacks/langfuse/langfuse.go`: Track startTime and compute duration

## Testing
Before fix: `Latency: 17751749h 37m 57s` ❌
After fix: `Latency: 854ms` ✅

---

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/callbacks/langfuse/go.mod
+++ b/callbacks/langfuse/go.mod
@@ -11,6 +11,8 @@ require (
 	github.com/stretchr/testify v1.10.0
 )
 
+replace github.com/cloudwego/eino-ext/libs/acl/langfuse => ../../libs/acl/langfuse
+
 require (
 	github.com/bahlo/generic-list-go v0.2.0 // indirect
 	github.com/buger/jsonparser v1.1.1 // indirect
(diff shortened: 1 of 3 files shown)


tangchaojun-bytedance:
@favbox hello, could you please solve conflicts?

tangchaojun-bytedance:
@favbox could you please check unit-test?

PR #695, opened 2026-02-26
refactor(ark): remove explicit tool choice assignment from chat model

JonXSnow:
Removed explicit `toolChoice` field assignment in tool binding methods for both standard ChatModel and ResponsesAPIChatModel.

Changes:
- Remove `toolChoice = &tc` assignment in WithTools(), BindTools(), and BindForcedTools() methods in chatmodel.go
- Remove `toolChoice` assignment in ResponsesAPIChatModel.WithTools() method in responses_api.go
- Simplify tool binding logic by relying on default behavior instead of explicit ToolChoiceAllowed/ToolChoiceForced settings

This change simplifies the codebase by removing redundant tool choice configuration, allowing the underlying implementation to handle tool choice behavior implicitly.


components/model/ark/chatmodel.go
```diff
@@ -497,10 +486,6 @@ func (cm *ChatModel) BindForcedTools(tools []*schema.ToolInfo) (err error) {
 		return err
 	}
 
-	tc := schema.ToolChoiceForced
```
hi-pender, line 500: BindForcedTools 方法中的，是不是不能删。 这个是明确的语义

PR #707, opened 2026-03-02
feat(backend): enhance backend with ripgrep support, background execution and code optimizations

JonXSnow:
- Replace manual file-walking grep with ripgrep (rg) for GrepRaw
  - Use --json output for structured match parsing
  - Support case-insensitive, multiline, file type and glob filtering

- Replace hand-rolled globToRegex with doublestar library for GlobInfo

- Refactor Read method
  - Return *filesystem.FileContent instead of raw string
  - Change offset to 1-based, increase default limit to 2000
  - Optimize line building: use WriteByte('\n') instead of string concatenation
  - Remove trailing newline from last line of output

- Add RunInBackendGround option for async command execution
  - Immediate response with background pipe draining
  - Cancellation support via context

- Refactor ExecuteStreaming into smaller focused functions:
  initStreamingCmd, runCmdInBackground, drainPipesConcurrently,
  streamCmdOutput, readStderrAsync, streamStdout, handleCmdCompletion

- Fix resource leak: close stdout pipe on StderrPipe/Start failure

- Add tests for background execution and updated Read behavior

### DIFF
--- a/components/model/ark/chatmodel.go
+++ b/components/model/ark/chatmodel.go
@@ -462,18 +462,12 @@ func (cm *ChatModel) WithTools(tools []*schema.ToolInfo) (fmodel.ToolCallingChat
 		return nil, fmt.Errorf("failed to convert to ark responsesAPI tools: %w", err)
 	}
 
-	tc := schema.ToolChoiceAllowed
-
 	ncm := *cm.chatModel
 	ncm.rawTools = tools
 	ncm.tools = arkTools
-	ncm.toolChoice = &tc
-
 	nrcm := *cm.respChatModel
 	nrcm.rawTools = tools
 	nrcm.tools = respTools
-	nrcm.toolChoice = &tc
-
 	return &ChatModel{
 		chatModel:     &ncm,
 		respChatModel: &nrcm,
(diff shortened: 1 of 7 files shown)


adk/backend/local/local.go
```diff
@@ -105,128 +107,159 @@ func (s *Local) LsInfo(ctx context.Context, req *filesystem.LsInfoRequest) ([]fi
...
+	lineNum := 1
 	linesRead := 0
 
-	for scanner.Scan() {
-		if lineIdx >= offset {
-			result.WriteString(fmt.Sprintf("%6d\t%s\n", lineIdx+1, scanner.Text()))
-			linesRead++
-			if linesRead >= limit {
-				break
+	for {
```
shentongmartin, line 147: 这个实现为什么改

adk/backend/local/local.go
```diff
@@ -105,128 +107,159 @@ func (s *Local) LsInfo(ctx context.Context, req *filesystem.LsInfoRequest) ([]fi
...
+			break
 		}
-		lineIdx++
 	}
 
-	if err := scanner.Err(); err != nil {
-		return "", fmt.Errorf("error reading file: %w", err)
-	}
+	return &filesystem.FileContent{
+		Content: strings.TrimSuffix(result.String(), "\n"),
```
shentongmartin, line 174: 这里为什么 trim suffix

adk/backend/local/local.go
JonXSnow: Scanner 有默认 64KB 的单行上限（MaxScanTokenSize），超长行会报 bufio.ErrTooLong

adk/backend/local/local.go
JonXSnow: 将最后一行的换行符 移除，不然有换行符，读取出来的会多一空行；

PR #752, opened 2026-03-31
optimize(openai): enhance APIError struct with InnerError field 

zbysir:
Enhance APIError struct with InnerError field.

- Add InnerError field to APIError for better error context
- Implement Unwrap method to facilitate error unwrapping in error handling
- Update convOrigAPIError function to populate InnerError with the original error

## Motivation
I found that the `convOrigAPIError` method leads to the loss of internal error information. Many providers (e.g., OpenRouter) customize their error messages rather than using only Code, Message, and Param. Therefore, preserving the original internal error is necessary.

An example:  
When using OpenRouter with OpenAI, an error occurs and returns a very brief but useless error message:  
`[NodeRunError] error, status code: 400, status: 400 Bad Request, message: Provider returned error.`

Developers cannot retrieve the full underlying error via `unwrap` or `error.As`.  

Regardless, wrapping an error while preserving the internal error is a better approach.


zbysir:
CI reported an error: Need to create a new tag\nThe following modules have changes and may need version updates:\n- components/model/openai (Current: components/model/openai/v0.1.10)\n\n\n⚠️ Please create and push new version tags for these modules after merging this PR.

Do I need to handle this?

tangchaojun-bytedance (approved):

PR #762, opened 2026-04-06
fix(gemini): set IncludeServerSideToolInvocations when using built-in tools with function calling

### LINKED ISSUE #761: Gemini: INVALID_ARGUMENT when using built-in tools (GoogleSearch, etc.) with function calling
**Describe the bug**

When using built-in tools (e.g. `GoogleSearch`, `URLContext`) alongside custom function calling tools, the Gemini API returns:

```
Error 400, Message: Please enable tool_config.include_server_side_tool_invocations to use Built-in tools with Function calling., Status: INVALID_ARGUMENT
```

The `GenerateContentConfig.ToolConfig` needs `IncludeServerSideToolInvocations` set to `true` when both built-in tools and function declarations are present.

**To Reproduce**

1. Create a ChatModel with `EnableGoogleSearch` configured
2. Pass function calling tools via `schema.ToolInfo`
3. Call `Generate` or `Stream`
4. API returns `INVALID_ARGUMENT`

**Expected behavior**

The adapter should automatically set `ToolConfig.IncludeServerSideToolInvocations = true` when any built-in tool (GoogleSearch, URLContext, CodeExecution, ComputerUse, FileSearch, GoogleSearchRetrieval, GoogleMaps) is enabled alongside function declarations.

**Version:**

gemini@v0.1.29

**Environment:**

google.golang.org/genai v1.51.0+ (field added in this version)

**Additional context**

The `IncludeServerSideToolInvocations` field was added to `genai.ToolConfig` in `google.golang.org/genai` v1.51.0. The current `go.mod` pins `v1.36.0` which predates this field, so the fix also requires bumping the genai dependency.

jredl-va:
#### What type of PR is this?

fix: A bug fix

#### Check the PR title.

- [ ] Attach the PR updating the user documentation if the current PR requires user awareness at the usage level.

en:
When built-in tools (GoogleSearch, URLContext, CodeExecution, ComputerUse, FileSearch, GoogleSearchRetrieval, GoogleMaps) are used alongside function calling tools, the Gemini API requires `ToolConfig.IncludeServerSideToolInvocations` to be set to `true`. Without this, the API returns `INVALID_ARGUMENT: Please enable tool_config.include_server_side_tool_invocations to use Built-in tools with Function calling.`

This PR:
1. Bumps `google.golang.org/genai` from v1.36.0 to v1.52.1 (the `IncludeServerSideToolInvocations` field was added in v1.51.0)
2. Automatically sets `IncludeServerSideToolInvocations = true` in the `ToolConfig` when any built-in tool is present alongside function declarations

Fixes #761


jredl-va:
cc @meguminnnnnnnnn @N3kox @JonXSnow — tagging for visibility as top contributors to the gemini component.

jredl-va:
@JonXSnow my apologies for the second direct ping, but I was hoping to get a review from you on this. This allows us to set labels on the vertex api calls for better telemetry. 

(please let me know if there is a contributors group / process for these tiny types of maintenance adjustments to conform to the changes in the vertex api)

tangchaojun-bytedance:
@jredl-va Thanks for the fix. I think this needs two more changes before merge.

1. When `include_server_side_tool_invocations` is enabled, `AUTO` mode is not supported. The Gemini docs say it should default to `VALIDATED`:
https://ai.google.dev/gemini-api/docs/tool-combination#limitations

Currently `WithTools` / `BindTools` can still produce `FunctionCallingConfigModeAuto`, so built-in tools + function declarations may still generate an invalid config. We should switch `AUTO` to `VALIDATED` in this case.

2. This flag is not supported by the Vertex backend. In `google.golang.org/genai@v1.52.1`, `toolConfigToVertex` rejects `includeServerSideToolInvocations`:
https://github.com/googleapis/go-genai/blob/v1.52.1/models.go#L3968-L3970

So we should return a clear unsupported error for `BackendVertexAI` instead of setting the flag unconditionally.

Could you also add regression tests for these two cases?

jredl-va:
@tangchaojun-bytedance thanks for the feedback! Both issues addressed with proper test coverage. 

- AUTO → VALIDATED coercion when built-in tools combine with function declarations (handles explicit AUTO, MODE_UNSPECIFIED, and the empty-string default).
- Vertex backend returns a clear up-front error instead of failing in the SDK.

jredl-va:
@mrh997 @shentongmartin sorry for the direct ping. This PR addressed @tangchaojun-bytedance's review back in June and has been waiting since. It's a small wiring change with tests and docs, and CI is green. Hoping to get this merged, or happy to reshape it if you'd prefer something similar.

PR #766, opened 2026-04-08
 feat(openrouter): add top-level cache control and support non-text parts 

JonXSnow:
## Summary

  Closes #763

  - Export `CacheControl` type and add `NewCacheControl` constructor for flexible cache control configuration
  - Add top-level `cache_control` support in request body, enabling automatic multi-turn caching for Anthropic Claude and explicit breakpoints for Gemini 2.5
  - Add `WithCacheControl` option for per-request cache control override
  - Remove text-only restriction for explicit cache control breakpoints, now supports image, document, tool_use, and tool_result content types
  - Simplify cache control logic in `buildRequestModifier` by removing redundant type checks
  - Fix `getMessageContentCacheControl` getter side effect on nil Extra

  ## Test plan

  - [x] Unit tests for top-level cache control with default and 1h TTL
  - [x] Unit tests for non-text part (image) cache control injection
  - [x] Unit tests for nil cache control option
  - [x] Existing cache control tests updated and passing
  - [x] `go test -race ./...` passes

### DIFF
--- a/components/model/openrouter/chatmodel.go
+++ b/components/model/openrouter/chatmodel.go
@@ -127,17 +127,27 @@ type Config struct {
 	// Keys are strings with a maximum length of 64 characters. Values are strings with a maximum length of 512 characters.
 	Metadata map[string]string `json:"metadata,omitempty"`
 
+	// CacheControl sets the top-level cache_control for all requests from this model instance.
+	// This enables automatic prompt caching for supported providers:
+	//   - Anthropic Claude: auto-caching (recommended for multi-turn conversations)
+	//   - Gemini 2.5: explicit breakpoints
+	// Can be overridden per-request via WithCacheControl option.
+	// Optional.
+	CacheControl *CacheControl `json:"cache_control,omitempty"`
+
 	// ExtraFields will override any existing fields with the same key.
 	// Optional. Useful for experimental features not yet officially supported.
 	ExtraFields map[string]any `json:"extra_fields,omitempty"`
 }
 
 type ChatModel struct {
-	cli            *openai.Client
+	cli *openai.Client
+
 	models         []string
 	reasoning      *Reasoning
 	responseFormat *ChatCompletionResponseFormat
 	metadata       map[string]string
+	cacheControl   *cacheControl
 }
 
 func NewChatModel(ctx context.Context, config *Config) (*ChatModel, error) {
@@ -189,6 +199,7 @@ func NewChatModel(ctx context.Context, config *Config) (*ChatModel, error) {
 	chatModel.reasoning = config.Reasoning
 	chatModel.metadata = config.Metadata
 	chatModel.responseFormat = config.ResponseFormat
+	chatModel.cacheControl = config.CacheControl.toInternal()
 
 	return chatModel, nil
 }
@@ -219,10 +230,18 @@ func (cm *ChatModel) buildRequestModifier(option *openrouterOption) openai.Reque
 			models          = option.models
 			reasoning       = option.reasoning
 			metadata        = option.metadata
+			responseFormat  = option.responseFormat
 			modifiedRawBody = make([]byte, len(rawBody))
 		)
 
 		_ = copy(modifiedRawBody, rawBody)
+
+		if responseFormat != nil {
+			modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, "response_format", responseFormat)
+			if err != nil {
+				return nil, err
+			}
+		}
 
 		if len(models) > 0 {
 			modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, "models", models)
@@ -230,6 +249,7 @@ func (cm *ChatModel) buildRequestModifier(option *openrouterOption) openai.Reque
 				return nil, err
 			}
 		}
+
 		if reasoning != nil {
 			modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, "reasoning", reasoning)
 			if err != nil {
@@ -237,15 +257,14 @@ func (cm *ChatModel) buildRequestModifier(option *openrouterOption) openai.Reque
 			}
 		}
 
-		if cm.metadata != nil {
+		if metadata != nil {
 			modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, "metadata", metadata)
 			if err != nil {
 				return nil, err
 			}
 		}
-
-		if cm.responseFormat != nil {
-			modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, "response_format", cm.responseFormat)
+		if option.cacheControl != nil {
+			modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, "cache_control", option.cacheControl)
 			if err != nil {
 				return nil, err
 			}
@@ -261,30 +280,25 @@ func (cm *ChatModel) buildRequestModifier(option *openrouterOption) openai.Reque
 
 			if len(msg.UserInputMultiContent) > 0 {
 				for cIdx, part := range msg.UserInputMultiContent {
-					if part.Type == schema.ChatMessagePartTypeText {
-						if ctrl, ok := getMessageInputPartCacheControl(&part); ok {
-							modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, fmt.Sprintf("messages.%d.content.%d.cache_control", index, cIdx), ctrl)
-							if err != nil {
-								return nil, err
-							}
-						}
-					} else {
-						return nil, fmt.Errorf("only 'ChatMessagePartTypeText' support cache control in 'MessageInputPart', but got part type '%s'", part.Type)
+					ctrl, ok := getMessageInputPartCacheControl(&part)
+					if !ok {
+						continue
+					}
+					modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, fmt.Sprintf("messages.%d.content.%d.cache_control", index, cIdx), ctrl)
+					if err != nil {
+						return nil, err
 					}
 				}
 
 			} else if len(msg.AssistantGenMultiContent) > 0 {
 				for cIdx, part := range msg.AssistantGenMultiContent {
-					if part.Type == schema.ChatMessagePartTypeText {
-						if ctrl, ok := getMessageOutputPartCacheControl(&part); ok {
-							modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, fmt.Sprintf("messages.%d.content.%d.cache_control", index, cIdx), ctrl)
-							if err != nil {
-								return nil, err
-							}
-						}
-					} else {
-						return nil, fmt.Errorf("only 'ChatMessagePartTypeText' support cache control in 'MessageOutputPart', but got part type '%s'", part.Type)
-
+					ctrl, ok := getMessageOutputPartCacheControl(&part)
+					if !ok {
+						continue
+					}
+					modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, fmt.Sprintf("messages.%d.content.%d.cache_control", index, cIdx), ctrl)
+					if err != nil {
+						return nil, err
 					}
 				}
 			} else if msg.Content != "" {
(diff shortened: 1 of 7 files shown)


components/model/openrouter/chatmodel.go
```diff
@@ -434,7 +450,15 @@ func (cm *ChatModel) WithTools(tools []*schema.ToolInfo) (model.ToolCallingChatM
...
 		return nil, err
 	}
-	return &ChatModel{cli: cli}, nil
+	return &ChatModel{
+		cli: cli,
+
+		models:         cm.models,
+		reasoning:      cm.reasoning,
+		responseFormat: cm.responseFormat,
+		metadata:       cm.metadata,
```
shentongmartin, line 460: 🔴 **Bug: Shallow-copy aliasing of `metadata` map**

`cm.metadata` is a `map[string]string` — assigning it here copies the *reference*, not the data. After `WithTools()`, both the original `ChatModel` and the new tool-enabled `ChatModel` share the **same underlying map**. If either side mutates the map (e.g. `cm.metadata["key"] = "new"`), the other is silently affected.

**Verified by test** `TestAttack_WithToolsMetadataAliasing` — mutating `cm.metadata` after `WithTools()` is reflected in the new model.

**Suggested fix:**
```go
metadata := make(map[string]string, len(cm.metadata))
for k, v := range cm.metadata {
    metadata[k] = v
}
// then use metadata: metadata,
```

components/model/openrouter/chatmodel.go
```diff
@@ -434,7 +450,15 @@ func (cm *ChatModel) WithTools(tools []*schema.ToolInfo) (model.ToolCallingChatM
 	if err != nil {
 		return nil, err
 	}
-	return &ChatModel{cli: cli}, nil
+	return &ChatModel{
+		cli: cli,
+
+		models:         cm.models,
```
shentongmartin, line 457: 🔴 **Bug: Shallow-copy aliasing of `models` slice**

`cm.models` is a `[]string` — assigning it copies the slice header (pointer, len, cap), so both the old and new `ChatModel` share the **same backing array**. Direct index mutation on the original (e.g. `cm.models[0] = "X"`) propagates to the tool model.

**Verified by test** `TestAttack_WithToolsModelsSliceAliasing` — mutating `cm.models[0]` after `WithTools()` is reflected in the new model.

**Suggested fix:**
```go
models := make([]string, len(cm.models))
copy(models, cm.models)
// then use models: models,
```

shentongmartin (commented):
🔍 **Attack-review**: 2 confirmed bugs found in `WithTools` — shallow-copy aliasing for `metadata` (map) and `models` (slice). See inline comments for details and suggested fixes.

components/model/openrouter/message_extra.go
```diff
@@ -129,14 +129,32 @@ const (
...
+type CacheControl struct {
+	TTL CacheControlTTL `json:"ttl,omitempty"`
+}
+
+func (c *CacheControl) toInternal() *cacheControl {
+	if c == nil {
+		return nil
+	}
+	cc := &cacheControl{Type: cacheControlEphemeralType, TTL: c.TTL}
+	if cc.TTL == "" {
```
shentongmartin, line 144: 🟡 **Design concern: No TTL validation**

Since `CacheControlTTL` is a `string` type, users can pass any arbitrary value like `CacheControl{TTL: "999y"}` and it will be sent verbatim to the OpenRouter API. The `toInternal()` method only defaults an *empty* TTL to `"5m"` but does not validate against the known constants (`CacheControlTTL5Minutes`, `CacheControlTTL1Hour`).

**Verified by test** `TestAttack_InvalidCacheControlTTL` — arbitrary TTL string passes through unchecked.

**Suggestion:** Consider adding validation here:
```go
switch cc.TTL {
case "", CacheControlTTL5Minutes:
    cc.TTL = CacheControlTTL5Minutes
case CacheControlTTL1Hour:
    // ok
default:
    // return error or log warning
}
```
Alternatively, document that only the two defined constants are supported.

components/model/openrouter/option.go
```diff
@@ -51,3 +53,20 @@ func WithMetadata(m map[string]string) model.Option {
...
+	return model.WrapImplSpecificOptFn(func(o *openrouterOption) {
+		o.cacheControl = ctrl.toInternal()
+	})
+}
+
+// WithResponseFormat sets the response format for the request.
+// When set, it overrides the ResponseFormat field configured in Config for this specific request.
+func WithResponseFormat(rf *ChatCompletionResponseFormat) model.Option {
+	return model.WrapImplSpecificOptFn(func(o *openrouterOption) {
+		o.responseFormat = rf
```
shentongmartin, line 71: 🟡 **Design concern: `WithResponseFormat(nil)` silently clears config-level setting**

Since `rf` is a `*ChatCompletionResponseFormat`, passing `nil` sets `o.responseFormat = nil`, which effectively **removes** the config-level `ResponseFormat` for that request. The `buildRequestModifier` checks `if responseFormat != nil` before serializing, so a nil override means no `response_format` in the request body at all.

**Verified by test** `TestAttack_WithResponseFormatNil` — config-level `json_object` format is dropped when `WithResponseFormat(nil)` is used.

This *might* be intentional ("clear the format for this request"), but it could surprise users. **Suggestion:** Either document this behavior in the godoc, or add a nil guard:
```go
func WithResponseFormat(rf *ChatCompletionResponseFormat) model.Option {
    return model.WrapImplSpecificOptFn(func(o *openrouterOption) {
        if rf != nil {
            o.responseFormat = rf
        }
    })
}
```

shentongmartin (commented):
🟡 **Attack-review follow-up**: 4 design concerns identified. These are not bugs but potential sources of user confusion. See inline comments.

components/model/openrouter/chatmodel.go
```diff
@@ -261,30 +280,25 @@ func (cm *ChatModel) buildRequestModifier(option *openrouterOption) openai.Reque
...
+					ctrl, ok := getMessageOutputPartCacheControl(&part)
+					if !ok {
+						continue
+					}
+					modifiedRawBody, err = sjson.SetBytes(modifiedRawBody, fmt.Sprintf("messages.%d.content.%d.cache_control", index, cIdx), ctrl)
+					if err != nil {
+						return nil, err
 					}
 				}
 			} else if msg.Content != "" {
```
shentongmartin, line 305: 🟡 **Design concern: `EnableMessageContentCacheControl` on empty `Content` is silently dropped**

If a user calls `EnableMessageContentCacheControl(msg)` on a message where `msg.Content == ""`, the cache control metadata is stored in `msg.Extra` but this branch is never entered — so the `cache_control` field is never written to the request body. This is a silent failure from the user's perspective.

**Verified by test** `TestAttack_EmptyContentWithCacheControl` — cache control on empty content message is silently ignored.

**Suggestion:** Consider either:
- Documenting that `EnableMessageContentCacheControl` requires non-empty `Content`
- Or logging a warning / returning an error in `EnableMessageContentCacheControl` when `msg.Content == ""`

components/model/openrouter/chatmodel.go
```diff
@@ -261,30 +280,25 @@ func (cm *ChatModel) buildRequestModifier(option *openrouterOption) openai.Reque
 
 			if len(msg.UserInputMultiContent) > 0 {
```
shentongmartin, line 282: 🟡 **Design concern: `UserInputMultiContent` takes priority — content-level cache control silently ignored**

When a message has both `UserInputMultiContent` (len > 0) and `Content` set, this `if` branch runs and the `else if msg.Content != ""` branch at line 304 is skipped. This means if a user calls `EnableMessageContentCacheControl(msg)` on such a message, the cache control is stored in `msg.Extra` but never serialized — it's silently lost.

**Verified by test** `TestAttack_MultiContentTakesPriorityOverContent` — content-level cache control is ignored when `UserInputMultiContent` is present.

This priority logic is correct by design, but could surprise users. **Suggestion:** Document this precedence in the godoc of `EnableMessageContentCacheControl`, e.g.:
> Note: This only applies when the message uses `Content` (string). If `UserInputMultiContent` is set, use `EnableMessageInputPartCacheControl` on individual parts instead.

components/model/openrouter/chatmodel.go
JonXSnow: 1. WithTools 内部只是共享，不修改：新旧 ChatModel 实例都只在 Generate 中只读使用 models 和 metadata，不会 append 或写入 map。                                                                                                                           
  2. config 生命周期短：NewChatModel 的调用方通常构造 config 后就不再持有或修改它，不存在"上层事后修改"的实际场景。
  3. 框架调用模式决定安全：WithTools 是框架在绑定工具时调用的，返回的实例由框架管理，用户代码不会同时操作原实例和新实例的这些字段

components/model/openrouter/option.go
JonXSnow: 设置为nil 应该是使用方决定的，如设置nil ，导致 ResponseFormat 失效，应该也属于合理范围

components/model/openrouter/chatmodel.go
shentongmartin: Thanks for the response. Your points are pragmatically reasonable:

1. ✅ **Read-only usage** — agreed, the current code only reads these fields in `Generate`/`Stream`.
2. ✅ **Short config lifecycle** — agreed, but note the aliasing is between the two `ChatModel` instances (original and tool-enabled), not between `Config` and `ChatModel`.
3. ⚠️ **Framework calling pattern** — this is an implementation-detail-level safety guarantee, not an API-contract-level one. If a future change adds mutation to these fields (e.g. middleware that appends to `models`), the bug would surface silently.

**Verdict:** I accept your reasoning for the current state — this is unlikely to be hit in practice. However, the cost of a deep copy is trivial (`make` + `copy`/range loop), while the cost of a future aliasing bug is high (silent data corruption). I'd still **recommend** the fix as defensive programming, but won't block on it.

Non-blocking 👍

components/model/openrouter/option.go
shentongmartin: Agreed ✅ — setting `nil` explicitly is a deliberate caller decision, and the behavior (clearing the config-level format) is consistent with standard Go option-override patterns. This is fine as-is.

Resolved.

components/model/openrouter/chatmodel.go
JonXSnow: 针对content 设置为空的场景，默认不作为模型的输入；cache control 也就不需要设置

components/model/openrouter/chatmodel.go
JonXSnow: UserInputMultiContent 与 content 优先使用 UserInputMultiContent

components/model/openrouter/message_extra.go
JonXSnow: 1. CacheControlTTL 是 string 类型常量：合法值只有 CacheControlTTL5Minutes 和 CacheControlTTL1Hour，都是包内定义的常量。用户传错值时，OpenRouter API 会返回错误，由服务端兜底。                                                                          
  2. 空值已有默认处理：145 行已经处理了 TTL 为空的情况，默认设为 5 分钟。                                                                                                                                                                                 
  3. 这是内部转换函数：toInternal() 只在 NewChatModel（202 行）和 WithCacheControl option 中调用，入口有限，不是公开的用户输入边界。

components/model/openrouter/chatmodel.go
shentongmartin: Agreed ✅ — empty `Content` means no actual text is sent to the model, so attaching cache control to it is a user error, not a framework bug. The behavior (silently skipping) is correct.

Resolved.

components/model/openrouter/chatmodel.go
shentongmartin: Agreed ✅ — this is the established precedence in the codebase: `UserInputMultiContent` > `AssistantGenMultiContent` > `Content`. Users who set `UserInputMultiContent` should use `EnableMessageInputPartCacheControl` on individual parts. The behavior is consistent and correct.

Resolved.

components/model/openrouter/message_extra.go
shentongmartin: Agreed ✅ — all three points are valid:

1. **Server-side fallback** — OpenRouter API will reject invalid TTL values, so the client doesn't need to duplicate that validation. The typed constants (`CacheControlTTL5Minutes`, `CacheControlTTL1Hour`) already guide users toward valid values.
2. **Empty default** — correctly handled at line 145.
3. **Limited entry points** — `toInternal()` is not a public user-input boundary, so the risk of garbage input is low.

The combination of typed constants + server-side validation is sufficient. No client-side validation needed.

Resolved.

shentongmartin (approved):

PR #798, opened 2026-04-24
feat(deepseek): add ThinkingConfig support for thinking/non-thinking mode

JonXSnow:
Add ThinkingConfig to ChatModelConfig to allow users to control the thinking mode switch. Upgrade deepseek-go from v1.3.2 to v1.3.4.


k8scat:
等着用

N3kox (approved):

PR #799, opened 2026-04-24
feat(deepseek): support ExtraFields via request-level option

jesse-engineer:
### What type of PR is this?

- [x] feat

### Summary

Introduce a new `WithExtraFields` request-level option for the DeepSeek chat model so callers can inject arbitrary passthrough fields into the top-level JSON payload sent to the DeepSeek API on a per-request basis.

This mirrors the `ExtraFields` (`extra_fields`) map that already exists on the underlying `deepseek-go` SDK's `ChatCompletionRequest` / `StreamChatCompletionRequest`, which the SDK merges into the outgoing request body via `finalizePayload`.

### Motivation

The DeepSeek API occasionally ships new request parameters (e.g. `chat_template_kwargs`, provider-specific flags) before this component — or even the underlying `deepseek-go` SDK — models them as first-class fields. Today, users of `components/model/deepseek` have no way to send those fields without maintaining a fork of this package.

`WithExtraFields` closes that gap in a minimal, forward-compatible way: arbitrary JSON keys can be attached to a single `Generate` / `Stream` call without touching `ChatModelConfig` or `ChatCompletionRequest`.

### Changes

- `components/model/deepseek/option.go` (new): define `deepseekOptions` and `WithExtraFields(map[string]interface{}) model.Option`.
- `components/model/deepseek/deepseek.go`:
  - `generateRequest`: read impl-specific options via `model.GetImplSpecificOptions` and forward `extraFields` to `deepseek.ChatCompletionRequest.ExtraFields`.
  - `generateStreamRequest`: propagate `ExtraFields` from the intermediate `ChatCompletionRequest` to `StreamChatCompletionRequest`.
- `components/model/deepseek/option_test.go` (new): unit tests covering:
  - option aggregation via `GetImplSpecificOptions`
  - nil map is a no-op
  - `Generate` forwards `ExtraFields` into the captured request
  - `Stream` forwards `ExtraFields` into the captured stream request
  - omitting the option leaves `ExtraFields` as `nil`

### Usage

```go
msg, err := cm.Generate(ctx, in,
    deepseek.WithExtraFields(map[string]interface{}{
        "chat_template_kwargs": map[string]interface{}{
            "thinking": true,
        },
    }),
)
```

### Backward compatibility

Purely additive. Existing callers are unaffected; when the option is not supplied, `ExtraFields` remains `nil` and the produced request is byte-identical to before.

### Tests

```bash
cd components/model/deepseek
go vet ./...
go build ./...
go test -gcflags="all=-l" -count=1 ./...
```

All existing and newly added tests pass locally.

### DIFF
--- a/components/model/deepseek/deepseek.go
+++ b/components/model/deepseek/deepseek.go
@@ -532,6 +532,7 @@ func (cm *ChatModel) generateStreamRequest(ctx context.Context, in []*schema.Mes
 		Tools:            origReq.Tools,
 		LogProbs:         origReq.LogProbs,
 		TopLogProbs:      origReq.TopLogProbs,
+		ExtraFields:      origReq.ExtraFields,
 	}
 	return req, cbIn, nil
 }
@@ -548,6 +549,8 @@ func (cm *ChatModel) generateRequest(_ context.Context, in []*schema.Message, op
 		ToolChoice:  cm.toolChoice,
 	}, opts...)
 
+	specOpts := model.GetImplSpecificOptions(&deepseekOptions{}, opts...)
+
 	req := &deepseek.ChatCompletionRequest{
 		Model:            *options.Model,
 		MaxTokens:        dereferenceOrZero(options.MaxTokens),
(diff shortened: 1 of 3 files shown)


jesse-engineer:
img_v3_02113_e46038af-0707-481c-94ad-e4c7134cee4g

支持deepseek的思考强度控制

shentongmartin (approved):

PR #808, opened 2026-04-28
[codex] fix deepseek thinking config for streams

### LINKED ISSUE #807: deepseek thinking config 没有应用
Image

k8scat:
## Summary
- forward DeepSeek ThinkingConfig into stream chat completion requests
- add regression coverage for stream requests carrying thinking config

## Root Cause
`generateStreamRequest` builds a stream request from the non-stream request but did not copy the `Thinking` field, so `ChatModelConfig.ThinkingConfig` was applied to Generate but dropped for Stream.

## Validation
- `git diff --check main...HEAD`
- `GOTOOLCHAIN=go1.24.1 go test ./... -count=1 -gcflags="all=-N -l"` in `components/model/deepseek`

Fixes #807


k8scat:
@JonXSnow 大佬帮忙review下

PR #812, opened 2026-04-29
feat(agentkit/local): add MultiModalRead for images and PDFs

JonXSnow:
Add MultiModalRead to SandboxTool that returns structured multimodal content for image and PDF files, with fallback to Read for other types.

Images:
- Read raw bytes via readAllBytes Python template (base64 transfer)
- Detect MIME type via magic-number headers (PNG/JPEG/GIF/BMP/WebP/TIFF)
- Enforce 10 MB size limit

PDFs:
- Full read (no 'pages'): return raw PDF bytes, 20 MB limit
- Paged read (with 'pages'): render page range to PNG via go-fitz at 150 DPI, 100 MB limit, max 20 pages per request
- Validate 'pages' parameter syntax and ranges upfront
- Clamp end page with warning when it exceeds total page count

Size enforcement:
- Dual-layer check: Python side (os.path.getsize) rejects before read, Go side re-checks after decode as defense-in-depth
- Use stdout sentinel marker (__READALLBYTES_TOO_LARGE__) instead of exit codes to distinguish size errors from other failures
- Provide actionable error messages (e.g. suggest 'pages' param when full-read limit is exceeded)

Also:
- Switch Write template to binary mode ('wb') for byte-level fidelity
- Add table-driven tests for parsePagesParam
- Bump go directive to 1.23; upgrade eino to v0.9.0-alpha.5


adk/backend/agentkit/sandbox.go
```diff
@@ -53,6 +56,96 @@ const (
...
+	// MaxPDFPagesPerRequest caps the number of pages rendered per paged read. Default 20. Hard-cap 1000.
+	MaxPDFPagesPerRequest int
+	// PDFRenderDPI is dots-per-inch used when rasterizing each PDF page to PNG.
+	// Higher DPI yields sharper images at the cost of larger payloads:
+	// typical screens are 72-96 DPI, 150 DPI ≈ 2x sharpness with manageable size,
+	// 300 DPI is print-grade but produces ~4x larger images.
+	// Default 150. Hard-cap 600.
+	PDFRenderDPI float64
+}
+
```
shentongmartin, line 100: **[suggestion] Code duplication (~230 LOC)**

The following types/functions are duplicated verbatim between `local/local.go` and `agentkit/sandbox.go`:

| Element | ~LOC |
|---------|------|
| `MultiModalReadConfig` + constants | 50 |
| `resolveMultiModalReadConfig`, `clampInt`, `clampFloat` | 30 |
| `parsePagesParam`, `splitPagesRange`, `validatePagesRange` | 65 |
| `renderPDFPagesToImages`, `newImageContentPart` | 25 |
| `isImageExt`, `isPDFExt`, `detectImageMIME`, `isPDFBytes` | 60 |
| `errReadAllBytesTooLarge` sentinel | 3 |

Consider extracting to `adk/backend/internal/multimodal` (or similar) to prevent the two copies from drifting over time. Tests are also duplicated correspondingly.

Non-blocking — manageable now, but will become a maintenance burden as features are added.

adk/backend/agentkit/sandbox.go
```diff
@@ -396,6 +499,349 @@ func (s *SandboxTool) Edit(ctx context.Context, req *filesystem.EditRequest) err
...
+	}
+
+	if paged {
+		if pagedStart > totalPages {
+			return nil, fmt.Errorf("invalid pages parameter: %q (start page %d exceeds total pages %d in file %s)", req.Pages, pagedStart, totalPages, path)
+		}
+		// Keep clamp to allow read-to-end style requests like "1-100" on short PDFs;
+		// surface a warn so the caller can notice the mismatch.
+		if pagedEnd > totalPages {
+			log.Printf("[WARN] MultiModalRead: end page %d exceeds total pages %d for %s (pages=%q), clamped to %d", pagedEnd, totalPages, path, req.Pages, totalPages)
```
shentongmartin: **[suggestion]** `log.Printf` in library code writes to stderr, which can surprise callers. Consider either:
- Returning the warning as a field in the response (e.g. `Warnings []string` in `MultiFileContent`)
- Accepting a `*slog.Logger` (or similar) via config

Same issue exists at `local/local.go:356`.

adk/backend/agentkit/code_template.go
```diff
@@ -20,8 +20,9 @@ const (
 	readPythonCodeTemplate = `
 import os
 import sys
+import base64
```
shentongmartin, line 23: **[suggestion] Pre-existing path injection inconsistency**

Good fix switching to `file_path_b64` here! However, the other templates in this file still use raw `'{path}'` interpolation:
- `lsInfoPythonCodeTemplate` (line ~80)
- `writePythonCodeTemplate`
- `editPythonCodeTemplate`

A file path containing a single quote would break those scripts or enable injection.

Not introduced by this PR, but the fix here makes the inconsistency more visible. May be worth unifying in a follow-up.

shentongmartin (commented):
Overall: high-quality PR, well-designed and defensively coded. 19 adversarial attack tests all pass clean — no confirmed bugs. A few suggestions for long-term maintainability below.

adk/backend/agentkit/sandbox.go
JonXSnow: agentkit 与 local 是两个独立的子mod ，所以最好的方式是各自维护

adk/backend/agentkit/code_template.go
JonXSnow: 已经将其他方法都统一修改成 base64 格式传递

shentongmartin (approved):

PR #817, opened 2026-05-03
feat(moonshot): add Moonshot (Kimi) ChatModel implementation

Narizis:
## Summary

Adds a new `components/model/moonshot` package that implements `model.ToolCallingChatModel` for [Moonshot AI (Kimi)](https://platform.moonshot.cn/docs/api/chat) via its OpenAI-compatible API.

The implementation is a thin wrapper over `libs/acl/openai` (same shape as `components/model/qwen`):

- Default `BaseURL` set to `https://api.moonshot.cn/v1`.
- `Generate` / `Stream` / `WithTools` / `BindTools` / `BindForcedTools` forwarded to the shared OpenAI client.
- Standard OpenAI parameters exposed in `ChatModelConfig` (model, max_tokens, temperature, top_p, stop, presence/frequency penalty, response_format, user).
- Moonshot-specific parameters (e.g. `partial` for prefix completion, `context_cache`) can be passed through `WithExtraFields`.

## Test plan

- [x] `go build ./...` and `go vet ./...` pass under Go 1.23 inside `components/model/moonshot/`.
- [x] End-to-end smoke test against the real Moonshot API (`moonshot-v1-8k`):
  - `Generate` returned a valid completion with usage info (33 prompt + 11 completion tokens).
  - `Stream` returned the same content via streaming chunks.
- [x] Tool-calling end-to-end against the real API (`moonshot-v1-8k`) with `WithTools`:
  - `Generate` returned a `tool_calls` entry with the correct function name and JSON arguments.
  - `Stream` chunks merged via `schema.ConcatMessages` produced the same `tool_calls` entry (no `Index=nil` workaround needed, unlike Qwen / DashScope).
  - Feeding a synthetic tool result back as a `schema.Tool` message produced the expected final assistant reply.


Narizis:
@shentongmartin @JonXSnow could you take a look when you have a chance? 
Thanks! 🙇

shentongmartin:
> @shentongmartin @JonXSnow could you take a look when you have a chance? Thanks! 🙇

@Narizis What's the difference between this and the OpenAI ChatModel?

tangchaojun-bytedance:
@Narizis hello, I noticed that you implemented it via it's OpenAI-compatible API. I see it is almost the same with openai model. can you explain the difference and the moonshot model implemention is important rather than use kimi by openai model due to it's OpenAI-compatible API ?

PR #819, opened 2026-05-07
feat(ark): support audio input in Responses API

JonXSnow:
- Add ChatMessagePartTypeAudioURL handling in convUserInputMultiContentToContentItems
- Add convMsgInputAudioToResponseContentItem to build ContentItemAudio from schema.MessageInputAudio (URL or Base64 + MIMEType)
- Bump volcengine-go-sdk from v1.2.9 to v1.2.26 for ContentItemAudio support
- Add unit tests covering audio URL / Base64 success paths and nil / missing-URL / missing-MIMEType error paths

### DIFF
--- a/components/model/ark/chat_completion_api.go
+++ b/components/model/ark/chat_completion_api.go
@@ -23,6 +23,7 @@ import (
 	"io"
 	"log"
 	"runtime/debug"
+	"strings"
 
 	"github.com/eino-contrib/jsonschema"
 	"github.com/volcengine/volcengine-go-sdk/service/arkruntime"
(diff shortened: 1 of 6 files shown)


components/model/ark/chat_completion_api.go
```diff
@@ -645,6 +656,28 @@ func (cm *completionAPIChatModel) toArkContent(msg *schema.Message) (*model.Chat
...
+				audioURL := &model.ChatMessageAudioURL{}
+				if part.Audio.URL != nil && *part.Audio.URL != "" {
+					audioURL.URL = *part.Audio.URL
+					audioURL.Format = audioFormatFromMIME(part.Audio.MIMEType)
+				} else if part.Audio.Base64Data != nil && *part.Audio.Base64Data != "" {
+					if part.Audio.MIMEType == "" {
+						return nil, fmt.Errorf("audio part must have MIMEType when using Base64Data")
+					}
+					audioURL.Data = *part.Audio.Base64Data
+					audioURL.Format = audioFormatFromMIME(part.Audio.MIMEType)
```
shentongmartin: **[Design Concern] Asymmetric `data:` prefix validation**

In the Responses API path, `convMessagePartCommonToURL` explicitly rejects Base64Data that starts with `data:` prefix (via `ensureDataURL`). However, here the Completion API audio path directly assigns `*part.Audio.Base64Data` to `audioURL.Data` without any such validation.

This means a user who accidentally passes `"data:audio/wav;base64,SGVsbG8="` instead of raw base64 will:
- Get a **clear error** in Responses API
- Get **silently corrupt data** in Completion API

Suggested fix:
```go
if strings.HasPrefix(*part.Audio.Base64Data, "data:") {
    return nil, fmt.Errorf("audio Base64Data must be raw base64, not a data URL")
}
```

This would bring parity with image/video paths which also reject `data:` prefix via `ensureDataURL`.

components/model/ark/chat_completion_api.go
```diff
@@ -428,6 +429,16 @@ func runeSlice2int64(in []rune) []int64 {
...
 
+// audioFormatFromMIME derives the ark SDK audio Format field from a MIME type.
+// Examples: "audio/mp3" -> "mp3", "audio/wav" -> "wav". Returns input as-is when
+// it is empty or does not carry the "audio/" prefix.
+func audioFormatFromMIME(mime string) string {
+	if mime == "" {
+		return ""
+	}
+	return strings.TrimPrefix(mime, "audio/")
+}
```
shentongmartin, line 445: **[Suggestion] Strip MIME parameters for robustness**

RFC 2045 allows MIME types to carry parameters (e.g., `"audio/mpeg; codecs=mp3"`). With the current implementation, `strings.TrimPrefix` will produce `"mpeg; codecs=mp3"` as the Format field, which is likely not a valid format string for the ark backend.

Suggested improvement:
```go
func audioFormatFromMIME(mime string) string {
    if mime == "" {
        return ""
    }
    mime = strings.TrimPrefix(mime, "audio/")
    if idx := strings.IndexByte(mime, ';'); idx >= 0 {
        mime = mime[:idx]
    }
    return strings.TrimSpace(mime)
}
```

Also, non-audio MIME types (e.g., `"video/mp4"`) pass through unmodified and become the Format field — which is semantically incorrect. Consider validating that the MIME starts with `audio/`.

components/model/ark/chat_completion_api_test.go
```diff
@@ -446,6 +446,61 @@ func TestCompletionAPIChatModel_toArkContent(t *testing.T) {
...
+
+			PatchConvey("Error on missing MIMEType for audio Base64Data", func() {
+				audioB64 := "SGVsbG9BdWRpbw=="
+				msg := &schema.Message{UserInputMultiContent: []schema.MessageInputPart{{Type: schema.ChatMessagePartTypeAudioURL, Audio: &schema.MessageInputAudio{MessagePartCommon: schema.MessagePartCommon{Base64Data: &audioB64}}}}}
+				_, err := cm.toArkContent(msg)
+				convey.So(err, convey.ShouldNotBeNil)
+			})
 		})
 
 		PatchConvey("AssistantGenMultiContent", func() {
```
shentongmartin: **[Test Quality] Error assertions could be stronger**

These error test cases only check `err != nil` without verifying the error message content. This means a regression that returns a *different* error (e.g., a nil-pointer panic caught elsewhere) would still pass.

Contrast with the Responses API tests which use `assert.ErrorContains(t, err, "audio field must not be nil")` — much more precise.

Suggested:
```go
convey.So(err.Error(), convey.ShouldContainSubstring, "must not be nil")
```

components/model/ark/responses_api.go
```diff
@@ -1481,6 +1490,25 @@ func convMsgInputVideoToResponseContentItem(video *schema.MessageInputVideo) (*r
...
+		AudioUrl: audioURL,
+	}
+
+	return &responses.ContentItem{
+		Union: &responses.ContentItem_Audio{
+			Audio: contentItemAudio,
+		},
+	}, nil
+
+}
```
shentongmartin, line 1510: **[Positive] Good reuse of `convMessagePartCommonToURL`**

This implementation correctly delegates to the shared `convMessagePartCommonToURL` helper, which provides robust validation (nil checks, `data:` prefix rejection, MIMEType requirement). This is cleaner and more consistent than the Completion API's inline approach.

The pattern follows `convMsgInputVideoToResponseContentItem` and `convMsgInputImageToResponseContentItem` exactly — nicely done.

shentongmartin (commented):
## PR #819 Review Summary

**Verdict: Approve with suggestions** — no blockers.

This is a clean, well-scoped PR that follows established patterns. The author correctly identified that the Completion API and Responses API require different handling for audio (separate fields vs. data URL) and implemented both paths consistently with the existing image/video patterns.

**Key findings (all suggestions, no blockers):**
1. Asymmetric `data:` prefix validation between Completion API (accepts silently) and Responses API (rejects)
2. `audioFormatFromMIME` doesn't strip MIME parameters (`audio/mpeg; codecs=mp3` -> `mpeg; codecs=mp3`)
3. Test assertions in Completion API tests could be stronger (check error message content)

See inline comments for details.

shentongmartin (approved):

PR #822, opened 2026-05-12
fix(agenticark): skip Caching field when no SessionCacheConfig set

JonXSnow:
- Introduce hasSessionCacheCfg flag in populateCache to track whether SessionCache is configured via instance Config or WithCache option
- Only assign responseReq.Caching when SessionCacheConfig is present; leave it nil otherwise instead of always sending Type=disabled
- Preserve existing behavior for Store, ExpireAt, PreviousResponseId and HeadPreviousResponseID handling

### DIFF
--- a/components/model/agenticark/model.go
+++ b/components/model/agenticark/model.go
@@ -695,14 +695,16 @@ func (m *Model) populateCache(in []*schema.AgenticMessage, responseReq *response
 	arkOpts *arkOptions) ([]*schema.AgenticMessage, error) {
 
 	var (
-		store       = false
-		enableCache = false
-		expireAtSec *int64
-		headRespID  *string
+		store              = false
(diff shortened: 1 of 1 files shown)


components/model/agenticark/model.go
```diff
@@ -697,12 +697,18 @@ func (m *Model) populateCache(in []*schema.AgenticMessage, responseReq *response
 	var (
 		store       = false
 		enableCache = false
-		expireAtSec *int64
-		headRespID  *string
+		// hasSessionCacheCfg indicates whether SessionCacheConfig is explicitly set via
```
mrh997: 注释删了吧

JonXSnow: done

mrh997 (approved):

PR #837, opened 2026-05-20
feat(gemini): add WithPrefixCacheTTL for CreatePrefixCache

jesse-engineer:
## Summary
- Add `gemini.WithPrefixCacheTTL(duration)` so callers can set prefix cache TTL per `CreatePrefixCache` invocation, similar to `openairesponse.WithPromptCacheRetention`.
- Extract `resolvePrefixCacheConfig` to merge call-time option with `Config.Cache` defaults (option takes precedence; clears `ExpireTime` when TTL option is set).
- Add unit test and update README.

## Motivation
Prefix cache TTL is often tied to a specific use case (e.g. chapter reviewer) rather than the shared chat model instance. Today TTL can only be configured at `NewChatModel` via `Config.Cache`, which forces all `CreatePrefixCache` calls on that model to share the same TTL.

## Test plan
- [x] `go test -run TestCreatePrefixCache ./components/model/gemini/...`

### DIFF
--- a/components/model/gemini/README.md
+++ b/components/model/gemini/README.md
@@ -220,7 +220,7 @@ type CacheConfig struct {
 
 This component supports two caching strategies to improve latency and reduce API calls:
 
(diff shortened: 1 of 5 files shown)


tangchaojun-bytedance:
thank you for your advice. we have solved this problem. thank you.

PR #842, opened 2026-05-23
feat(acp): backend improvements

fanlv:
## What

Backend & streaming improvements for the `acp` package.

### `acp/backend.go`
- Introduce `ACPConn` interface so callers can substitute the connection (testing/proxying); `*conn.AgentConnection` still satisfies it.
- Add sentinel errors (`ErrShellNonZeroExit`, `ErrCapabilityMissing`, `ErrOldStringNotFound`, `ErrAmbiguousReplace`, `ErrFileTooLarge`) for structured error handling.
- Add `Config.validate()` and require `Capabilities`; clean up middleware wiring.
- Edit tool no longer requires terminal — it always uses fs `ReadTextFile` + `WriteTextFile` and rejects files beyond `maxEditFileSize` (32 MiB) to prevent OOM.
- Grep prefers `rg --json` and falls back transparently to POSIX `grep -RnE` when ripgrep is unavailable.
- Properly release terminals via a background context with timeout so cancellation does not leak terminal handles.
- Optional `slog.Logger` for non-fatal diagnostics.

### `acp/conv.go`
- `fromToolCall` now returns an error on invalid JSON arguments instead of silently coercing them.
- Streaming path flushes any tool calls whose arguments are already complete JSON before yielding the stream error, so a single broken tool call no longer drops the valid ones.

### Tests
- New `acp/backend_test.go`.
- Added `TestFromToolCallEmptyArgs`, `TestFromToolCallInvalidJSON`, and `TestAgentEventToSessionUpdate_StreamingFlushDoesNotDropValidToolCalls` in `conv_test.go`.

### Misc
- Example agent: concatenate all text blocks from prompts and surface stream errors to stderr.
- Update `acp/README.md` / `acp/README_zh.md` for the new behavior.
- `.gitignore`: ignore `*.local.md`.

## Why

- Make the ACP backend more robust against malformed model output (invalid JSON tool args), large files, and missing client tools (`rg`).
- Make the package easier to test by depending on an interface instead of the concrete connection type.
- Avoid resource leaks on cancelled requests.

## Test
- `go test ./acp/...`

### DIFF
--- a/acp/backend.go
+++ b/acp/backend.go
@@ -66,7 +66,7 @@ const (
 	// in memory. Files larger than this are rejected to prevent OOM.
 	maxEditFileSize = 32 * 1024 * 1024 // 32 MiB
 
-	// rgJSONMaxFailedRatio is the maximum ratio of unparseable lines in `rg --json`
+	// rgJSONMaxFailedRatio is the maximum ratio of unparsable lines in `rg --json`
 	// output before we treat the entire result as an error. This guards against
 	// incompatible rg versions that emit non-JSON output while still tolerating
 	// occasional malformed lines (e.g. from binary file matches or locale issues).
@@ -844,7 +844,7 @@ func parsePosixGrepOutput(out string) []filesystem.GrepMatch {
 // substring matching :<positive-integer><sep> before the real line number.
 //
 // Known limitation: if the path itself contains a `:N:` or `:N-` pattern where
-// N is a positive integer (e.g. "prefix:1:rest.go"), the parser will mis-split.
+// N is a positive integer (e.g. "prefix:1:rest.go"), the parser will split incorrectly.
 // POSIX grep has no NUL-delimited output mode (`-Z` is GNU-only), so this is
 // an inherent best-effort trade-off.
 func parsePosixGrepLine(line string) (filePath string, lineno int, content string, ok bool) {
--- a/acp/backend.go
+++ b/acp/backend.go
@@ -18,43 +18,120 @@ package einoacp
 
 import (
 	"context"
-	"encoding/base64"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"log/slog"
+	"path"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+	"unicode/utf8"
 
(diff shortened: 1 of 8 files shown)


acp/backend.go
```diff
@@ -18,43 +18,120 @@ package einoacp
...
+	// incompatible rg versions that emit non-JSON output while still tolerating
+	// occasional malformed lines (e.g. from binary file matches or locale issues).
+	rgJSONMaxFailedRatio = 0.5
+)
+
+// ACPConn is the minimal interface for ACP client-side RPC calls used by
+// Backend and shell. It is satisfied by *acpconn.AgentConnection (from the
+// github.com/eino-contrib/acp/conn package). Callers may provide alternative
+// implementations for testing or proxying.
+type ACPConn interface {
```
hi-pender, line 80: 这个是专门用来暴露 ACP Client Tool 的接口么？  还是代表 ACP Client 侧的 Conn 时，如果是后者的话，当前暴露的方法不足

acp/backend.go
```diff
@@ -171,13 +259,69 @@ func (s *shell) Execute(ctx context.Context, input *filesystem.ExecuteRequest) (
...
+	rgState atomic.Int32
+
+	// rgProbeMu serializes concurrent probe attempts so only one RPC flies at
+	// a time. Goroutines that can't acquire it immediately return false (POSIX
+	// fallback) rather than blocking.
+	rgProbeMu sync.Mutex
 }
 
-func (b *backend) Read(ctx context.Context, req *filesystem.ReadRequest) (*filesystem.FileContent, error) {
+// NewBackend constructs a Backend from the same Config used by
```
hi-pender, line 289: 这一行只有一半

hi-pender (approved):

acp/backend.go
fanlv: 只暴露了，跟 Backend 相关的方法。最小依赖原则。

acp/backend.go
fanlv: // NewBackend constructs a Backend from the same Config used by NewClientToolsMiddleware. 最后一个单测下一行。自动分开的，都是 AI 生成的。

PR #565, opened 2025-11-24
fix(gemini): Merge Tool Messages into Content Parts

yesiyuandy-bytedance:
#### What type of PR is this?

Summary

  - Implement proper tool message to Gemini FunctionResponse part conversion
  - Add logic to merge consecutive tool response messages into a single Content with multiple
  Parts, matching Gemini API requirements
  - Enhance thought_signature handling to support both []byte and base64-encoded string types,
  enabling JSON serialization round-trip

  Changes

  gemini.go

  - Add new convToolMessageToPart function to convert tool response messages to Gemini
  FunctionResponse parts
  - Refactor convSchemaMessages to detect and merge consecutive tool messages into a single
  user-role Content
  - Refactor convSchemaMessage to use the new tool message conversion function
  - Simplify control flow by moving tool message handling to the top of the function

  message_extra.go

  - Enhance getThoughtSignature to handle both []byte and base64-encoded string types
  - This fixes issues where thought_signature was lost after JSON marshaling/unmarshaling

#### Check the PR title.

- [ ] This PR title match the format: \<type\>(optional scope): \<description\>
- [ ] The description of this PR title is user-oriented and clear enough for others to understand.
- [ ] Attach the PR updating the user documentation if the current PR requires user awareness at the usage level. [User docs repo](https://github.com/cloudwego/cloudwego.github.io)

PR #872, opened 2026-06-08
fix(claude): merge consecutive tool messages into single UserMessage

mjlzz:
## Problem

When an assistant invokes multiple tools in a single turn, Eino produces consecutive `schema.Tool` messages. The current `populateInput` converts each one to a separate `UserMessage` via `convSchemaMessage`, but the Anthropic API requires all `tool_result` blocks for the same assistant turn to be delivered in a **single** user message.

This results in a 400 error from the Anthropic API when the agent makes parallel tool calls.

## Root Cause

The message loop in `populateInput` uses a simple `for _, msg := range msgs` loop that converts each message independently. For Tool-role messages, each becomes a separate Anthropic `UserMessage` — but consecutive user messages violate the Anthropic message structure requirement.

## Fix

Detect consecutive Tool-role messages in `populateInput`, collect them, convert each via `convSchemaMessage`, and merge all resulting content blocks into one `NewUserMessage`.

This is the same approach used to fix the equivalent issue for Gemini in #565.

## Changes

- `claude.go`: +32/−1 in `populateInput` to merge consecutive tool messages
- `claude_test.go`: +108 lines, 6 sub-tests covering:
  - Single tool message (no merge)
  - Two consecutive tools (merged into 1 message, 2 content blocks)
  - Three consecutive tools (merged into 1 message, 3 content blocks)
  - Tools separated by non-tool message (not merged)
  - No tool messages (passthrough)
  - Real-world multi-tool-call pattern

### DIFF
--- a/components/model/claude/claude.go
+++ b/components/model/claude/claude.go
@@ -652,7 +652,37 @@ func (cm *ChatModel) populateInput(params *anthropic.MessageNewParams, system []
 	msgParams := make([]anthropic.MessageParam, 0, len(msgs))
 	hasSetMsgBreakPoint := false
 
-	for _, msg := range msgs {
+	for i := 0; i < len(msgs); {
+		msg := msgs[i]
+
+		// Merge consecutive tool messages into a single UserMessage.
+		// Anthropic API requires all tool_result blocks for a given
+		// assistant turn to be in the same user message.
+		if msg.Role == schema.Tool {
+			// Collect all consecutive tool messages
+			toolMsgs := []*schema.Message{msg}
+			j := i + 1
+			for j < len(msgs) && msgs[j].Role == schema.Tool {
+				toolMsgs = append(toolMsgs, msgs[j])
+				j++
+			}
+
+			// Convert each and merge content blocks into one UserMessage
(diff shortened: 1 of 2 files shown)


mrh997:
@mjlzz [but consecutive user messages violate the Anthropic message structure requirement.] What error was reported ?

mrh997 (approved):

mrh997:
As a follow-up idea (not blocking this PR): the merge could also be done at the SDK layer — keep the conversion loop a pure 1:1 mapping and run a single normalization pass over the resulting `[]anthropic.MessageParam`. This keeps the wire-format constraint in one place, consistent with the gemini chat model (#565).

See #876 for a reference implementation of this approach (same fix for the agentic claude/gemini models).

mrh997:
sry, I accidentally merged it.

mjlzz:
@mrh997 你的建议比较合理，已合并的代码是撤掉还是由你提交新修改覆盖掉

mrh997:
@mjlzz 感谢你的 PR，我已经在另一个 PR 上进行了修改并发布了 v.0.1.20 。

PR #875, opened 2026-06-08
fix(agentic): tolerate unknown stream variants instead of failing

mrh997:
## Summary

The agenticopenai and agenticark Responses stream convertors previously returned an error the moment they encountered an unrecognized stream event, output item, or content part type, which aborted the entire stream. Providers add new event/item/part variants over time, so a single unknown variant could break otherwise-valid streaming responses. These convertors now skip unknown variants and continue.

## Key Changes

1. Main stream dispatch (`receivedStreamingResponse` / `receivedStreamResponse`): the `default` branch now `continue`s instead of emitting an `unknown event type` error. The explicit no-op `case` lists are dropped since they collapse into the lenient default.
2. `itemAddedEventToContentBlock` / `itemDoneEventToContentBlocks` (ark) and the content-part handlers (`contentPartAddedEventToContentBlock`, `contentPartDoneEventToContentBlock`, `eventContentPartToContentBlock`): unknown item / content part types now return `(nil, nil)` instead of erroring. Internal-invariant switches over already-built `block.Type` still error.
3. `sendBlock` now skips sending when both block and error are nil, so the ignored variants do not produce empty content blocks.
4. agenticopenai content-part handlers were restructured into a single type switch; the done-event handler also accepts `ResponseOutputRefusal` (it previously only matched `ResponseOutputText`).

### DIFF
--- a/components/model/agenticopenai/responses_event_convertor.go
+++ b/components/model/agenticopenai/responses_event_convertor.go
@@ -44,18 +44,6 @@ func receivedStreamingResponse(sr *ssestream.Stream[responses.ResponseStreamEven
 		sender.errHeader = fmt.Sprintf("failed to convert event '%s'", event.Type)
 
 		switch variant := event.AsAny().(type) {
-		case responses.ResponseTextDoneEvent,
-			responses.ResponseReasoningSummaryPartAddedEvent,
-			responses.ResponseReasoningSummaryPartDoneEvent,
-			responses.ResponseReasoningSummaryTextDoneEvent,
-			responses.ResponseFunctionCallArgumentsDoneEvent,
-			responses.ResponseMcpCallArgumentsDoneEvent,
-			responses.ResponseRefusalDoneEvent,
-			responses.ResponseCodeInterpreterCallCodeDoneEvent:
-
-			// Do nothing.
-			continue
-
 		case responses.ResponseErrorEvent:
 			_ = sw.Send(nil, fmt.Errorf("received error event: code=%s message=%s", variant.Code, variant.Message))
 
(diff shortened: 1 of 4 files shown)


components/model/agenticopenai/responses_event_convertor.go
```diff
@@ -208,7 +196,8 @@ func receivedStreamingResponse(sr *ssestream.Stream[responses.ResponseStreamEven
 			sender.sendBlock(block, nil)
 
 		default:
-			sw.Send(nil, fmt.Errorf("unknown event type: %s", event.Type))
+			// Unhandled event types are ignored rather than failing the stream.
+			continue
```
shentongmartin, line 200: 这里 top-level event loop 的 default 改成了静默丢弃，但 `itemAddedEventToContentBlock`（L398）和 `itemDoneEventToContentBlocks`（L573）的 default 分支还是 `return nil, fmt.Errorf("unknown item type...")`。

这两个方法的 error 会通过上面各 case 分支里的 `sender.sendBlock(block, err)` 传播到流上（因为 `sendBlock` 的 nil guard 只拦 `block==nil && err==nil` 的情况，不拦 `err!=nil`），所以如果 OpenAI 新增 item type，流仍然会挂。

对比 `agenticark` 里 `itemAddedEventToContentBlock` 和 `itemDoneEventToContentBlocks` 的 default 已经改成了 `return nil, nil`，建议这里也对齐：

```go
default:
    // Unknown item types are ignored rather than failing the stream.
    return nil, nil
```

mrh997: done

shentongmartin (approved):
LGTM. Tolerance pattern is now consistently applied at all levels in both packages.

PR #888, opened 2026-06-16
feat(claude): support Vertex service account JSON credentials

### LINKED ISSUE #890: feat(claude): support explicit Vertex service account JSON credentials
## Problem

`components/model/claude` supports multiple deployment backends, but Vertex AI authentication is inconsistent with the other backends.

| Backend | Explicit in-memory credentials | Fallback |
| --- | --- | --- |
| Direct Anthropic API | `APIKey` / `AuthToken` | SDK default credential discovery |
| AWS Bedrock | `AccessKey` / `SecretAccessKey` / `SessionToken` | AWS shared config / profile |
| Google Vertex AI | **Not supported** | Application Default Credentials (ADC) only |

When `ByVertex` is enabled, `NewChatModel` always calls `vertex.WithGoogleAuth`, which relies on ADC (e.g. `GOOGLE_APPLICATION_CREDENTIALS`, gcloud user credentials, or GCE/GKE metadata).

This is a problem for callers who:

- Load GCP service account JSON from a secret manager or config service at runtime
- Run in environments where ADC is unavailable or undesirable (local dev, CI, multi-tenant apps)
- Want parity with the Anthropic Go SDK, which already exposes `vertex.WithCredentials` for explicit credentials

Today, the only workaround is to write the JSON to disk and point `GOOGLE_APPLICATION_CREDENTIALS` at it — awkward for in-memory secrets and inconsistent with how Bedrock credentials are handled on `Config`.

## Proposed solution

Add `Config.VertexServiceAccountJSON` (`[]byte`):

- **Non-empty:** build credentials with `google.CredentialsFromJSON` and authenticate via `vertex.WithCredentials`
- **Empty:** keep existing ADC behavior via `vertex.WithGoogleAuth` (no breaking change)

## Related PR

- #888

shaikustin:
#### What type of PR is this?

feat

#### Check the PR title.

feat(claude): 支持通过 Vertex 服务账号 JSON 配置凭证

en:
`components/model/claude` already supports explicit in-memory credentials for direct Anthropic API (`APIKey` / `AuthToken`) and AWS Bedrock (`AccessKey` / `SecretAccessKey`). Vertex was ADC-only via `vertex.WithGoogleAuth`.

This PR adds `Config.VertexServiceAccountJSON` so callers can pass raw GCP service account JSON and authenticate with `vertex.WithCredentials`, without Application Default Credentials or environment variables.

- When `VertexServiceAccountJSON` is non-empty: build credentials via `google.CredentialsFromJSON` and call `vertex.WithCredentials`.
- When empty: existing ADC behavior is unchanged (`vertex.WithGoogleAuth`).
- Mirrors the Bedrock credential pattern on `Config`.
- Adds `TestVertexServiceAccountJSON` (invalid JSON, ADC fallback, mocked explicit-credential success path).
- Documents the new field and auth rules in `README.md`, `README_zh.md`, and `examples/generate`.

zh:
`components/model/claude` 已支持直连 Anthropic（`APIKey` / `AuthToken`）和 AWS Bedrock（`AccessKey` / `SecretAccessKey`）的内存凭证；Vertex 此前仅支持通过 `vertex.WithGoogleAuth` 使用 ADC。

本 PR 新增 `Config.VertexServiceAccountJSON`，允许调用方传入原始 GCP service account JSON，并通过 `vertex.WithCredentials` 鉴权，无需 ADC 或环境变量。

- `VertexServiceAccountJSON` 非空：通过 `google.CredentialsFromJSON` 构建凭证并调用 `vertex.WithCredentials`。
- 为空：保持现有 ADC 行为（`vertex.WithGoogleAuth`）。
- 与 Bedrock 在 `Config` 上显式传凭证的模式一致。
- 新增 `TestVertexServiceAccountJSON`（非法 JSON、ADC 回退、mockey 覆盖的显式凭证成功路径）。
- 在 `README.md`、`README_zh.md` 和 `examples/generate` 中补充字段与鉴权说明。

Fixes #890

#### (optional) The PR that updates user documentation:

Updated component README files in this PR.

### DIFF
--- a/components/model/claude/claude.go
+++ b/components/model/claude/claude.go
@@ -83,7 +83,7 @@ func NewChatModel(ctx context.Context, config *Config) (*ChatModel, error) {
 			return nil, errors.New("ByVertex is true but no region provided; set VertexRegion or CLOUD_ML_REGION")
 		}
 		if len(config.VertexServiceAccountJSON) > 0 {
-			googleCreds, err := google.CredentialsFromJSONWithType(ctx, config.VertexServiceAccountJSON, google.ServiceAccount,
+			googleCreds, err := google.CredentialsFromJSON(ctx, config.VertexServiceAccountJSON,
 				"https://www.googleapis.com/auth/cloud-platform")
 			if err != nil {
 				return nil, fmt.Errorf("create vertex credentials from service account JSON: %w", err)
--- a/components/model/claude/claude.go
+++ b/components/model/claude/claude.go
@@ -83,6 +83,9 @@ func NewChatModel(ctx context.Context, config *Config) (*ChatModel, error) {
 			return nil, errors.New("ByVertex is true but no region provided; set VertexRegion or CLOUD_ML_REGION")
 		}
 		if len(config.VertexServiceAccountJSON) > 0 {
+			// CredentialsFromJSON requires an explicit scope; cloud-platform covers Vertex AI.
+			// The ADC path (WithGoogleAuth) omits scopes and lets FindDefaultCredentials
+			// resolve them from the runtime environment instead.
 			googleCreds, err := google.CredentialsFromJSON(ctx, config.VertexServiceAccountJSON,
 				"https://www.googleapis.com/auth/cloud-platform")
 			if err != nil {
--- a/components/model/claude/claude.go
+++ b/components/model/claude/claude.go
@@ -34,6 +34,7 @@ import (
 	"github.com/anthropics/anthropic-sdk-go/vertex"
 	awsConfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/credentials"
+	"golang.org/x/oauth2/google"
 	"github.com/eino-contrib/jsonschema"
 
 	"github.com/cloudwego/eino/components"
@@ -81,7 +82,16 @@ func NewChatModel(ctx context.Context, config *Config) (*ChatModel, error) {
 		if region == "" {
 			return nil, errors.New("ByVertex is true but no region provided; set VertexRegion or CLOUD_ML_REGION")
 		}
-		cli = anthropic.NewClient(vertex.WithGoogleAuth(ctx, region, projectID))
+		if len(config.VertexServiceAccountJSON) > 0 {
+			googleCreds, err := google.CredentialsFromJSONWithType(ctx, config.VertexServiceAccountJSON, google.ServiceAccount,
+				"https://www.googleapis.com/auth/cloud-platform")
+			if err != nil {
+				return nil, fmt.Errorf("create vertex credentials from service account JSON: %w", err)
+			}
+			cli = anthropic.NewClient(vertex.WithCredentials(ctx, region, projectID, googleCreds))
+		} else {
+			cli = anthropic.NewClient(vertex.WithGoogleAuth(ctx, region, projectID))
+		}
 	} else if config.ByBedrock {
 		// Use AWS Bedrock
 		var opts []func(*awsConfig.LoadOptions) error
@@ -204,6 +214,12 @@ type Config struct {
 	// If not set, auto-detected from CLOUD_ML_REGION environment variable.
 	// See: https://claude.ai/docs/en/google-vertex-ai
 	VertexRegion string
+
+	// VertexServiceAccountJSON is raw GCP service account JSON for Vertex.
+	// When non-empty, credentials are built in-memory and passed to vertex.WithCredentials.
+	// When empty and ByVertex is true, vertex.WithGoogleAuth (ADC) is used instead.
+	// Optional for Vertex.
+	VertexServiceAccountJSON []byte
 
 	// BaseURL is the custom API endpoint URL
 	// Use this to specify a different API endpoint, e.g., for proxies or enterprise setups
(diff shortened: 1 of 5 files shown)


components/model/claude/claude.go
```diff
@@ -80,7 +81,16 @@ func NewChatModel(ctx context.Context, config *Config) (*ChatModel, error) {
 		if region == "" {
 			return nil, errors.New("ByVertex is true but no region provided; set VertexRegion or CLOUD_ML_REGION")
 		}
-		cli = anthropic.NewClient(vertex.WithGoogleAuth(ctx, region, projectID))
+		if len(config.VertexServiceAccountJSON) > 0 {
+			googleCreds, err := google.CredentialsFromJSON(ctx, config.VertexServiceAccountJSON,
+				"https://www.googleapis.com/auth/cloud-platform")
```
shentongmartin, line 90: nit: Consider adding a brief comment explaining why the `cloud-platform` scope is used here. Future maintainers may wonder whether this scope could be narrowed (e.g., to a more specific Vertex AI scope). A one-line note like `// cloud-platform scope is required for Vertex AI API access` would help clarify the intent.

Note that the ADC path (via `vertex.WithGoogleAuth`) doesn't pass any explicit scopes and relies on defaults — there's a subtle asymmetry here. If that's intentional, a short comment would be valuable.

components/model/claude/claude_test.go
```diff
@@ -1162,3 +1165,47 @@ func TestPopulateInputMergeConsecutiveTools(t *testing.T) {
...
+		assert.ErrorContains(t, err, "create vertex credentials from service account JSON")
+	})
+
+	t.Run("ADC path when service account JSON empty", func(t *testing.T) {
+		mockey.PatchConvey("", t, func() {
+			mockey.Mock(vertex.WithGoogleAuth).Return(option.WithAPIKey("adc-test")).Build()
+
+			model, err := NewChatModel(ctx, base)
+			assert.NoError(t, err)
+			assert.NotNil(t, model)
```
shentongmartin, line 1244: nit: These assertions verify no error and non-nil model, but they don't actually verify that `vertex.WithGoogleAuth` was called (as opposed to `vertex.WithCredentials`). If someone accidentally swapped the if/else branches, both mock tests would still pass since both mocks return a valid option.

Consider using a call-tracking pattern to verify the expected function was actually invoked:

```go
adcCalled := false
mockey.Mock(vertex.WithGoogleAuth).To(func(ctx context.Context, region string, projectID string, scopes ...string) option.RequestOption {
    adcCalled = true
    return option.WithAPIKey("adc-test")
}).Build()
// ...
assert.True(t, adcCalled)
```

components/model/claude/claude_test.go
```diff
@@ -1162,3 +1165,47 @@ func TestPopulateInputMergeConsecutiveTools(t *testing.T) {
...
+			mockey.Mock(google.CredentialsFromJSON).Return(&google.Credentials{
+				ProjectID: "from-sa",
+			}, nil).Build()
+			mockey.Mock(vertex.WithCredentials).Return(option.WithAPIKey("sa-test")).Build()
+
+			cfg := *base
+			cfg.VertexServiceAccountJSON = []byte(`{"type":"service_account","project_id":"from-sa"}`)
+			model, err := NewChatModel(ctx, &cfg)
+			assert.NoError(t, err)
+			assert.NotNil(t, model)
```
shentongmartin, line 1267: nit: Same as the ADC test above — this test doesn't verify that `vertex.WithCredentials` (and `google.CredentialsFromJSON`) were actually called with the right arguments. Recommend adding call-verification assertions so future refactors don't accidentally break the branch selection logic.

For extra robustness, you could also verify the projectID and region are forwarded correctly to `vertex.WithCredentials`.

components/model/claude/claude_test.go
```diff
@@ -1168,3 +1171,66 @@ func TestPopulateInputMergeConsecutiveTools(t *testing.T) {
...
+			assert.True(t, calledWithGoogleAuth, "expected vertex.WithGoogleAuth to be called for ADC path")
+		})
+	})
+
+	t.Run("service account JSON uses WithCredentials path", func(t *testing.T) {
+		mockey.PatchConvey("", t, func() {
+			var calledCredentialsFromJSON bool
+			var calledWithCredentials bool
+			mockey.Mock(google.CredentialsFromJSON).To(func(ctx context.Context, jsonData []byte, scopes ...string) (*google.Credentials, error) {
```
shentongmartin: This test is **currently failing** — the `google.CredentialsFromJSON` mock with `.To()` doesn't work with mockey. The real function gets called instead of the mock, so `calledCredentialsFromJSON` stays false and the assertion fails:

```
Error: Should be true
Messages: expected google.CredentialsFromJSON to be called for service account path
```

I ran into the same issue during the initial review. `mockey.Mock(google.CredentialsFromJSON).To(func(...) {...})` silently fails to hook this specific function.

**Suggested fix**: Use `.Return()` for `google.CredentialsFromJSON` (the way it was in the original version) and keep the `.To()` call-verification for `vertex.WithCredentials` only. Since `vertex.WithCredentials` requires the creds object that `CredentialsFromJSON` produces, if `WithCredentials` was called with the right project/region, you know the SA JSON path was taken correctly. That still gives you call-verification without relying on a mock that doesn't work for this particular function.

shentongmartin (changes requested):

components/model/claude/claude_test.go
shaikustin: good catch, fixed

shentongmartin (approved):

shentongmartin (approved):

PR #895, opened 2026-06-24
fix(claude): collapse invalid tool_call Arguments to empty object (not just empty)

### LINKED ISSUE #894: fix(claude): convToolCalls crashes with "unexpected end of JSON input" when tool_call Arguments is invalid (not just empty)
**Describe the bug**

`convToolCalls` in `components/model/claude/claude.go` only guards tool-call `Arguments` against the **empty** case before wrapping it in `json.RawMessage`. When a model emits a tool call whose `Arguments` is **non-empty but invalid JSON** — most commonly a *truncated* tool_use streamed for a large payload — the raw partial JSON is passed straight to `anthropic.NewToolUseBlock`, and `ToolUseBlockParam.MarshalJSON` then fails when the message is re-assembled for the next request:

```
[NodeRunError] create new streaming message fail: json: error calling MarshalJSON
for type anthropic.MessageParam: json: error calling MarshalJSON for type
anthropic.ContentBlockParamUnion: json: error calling MarshalJSON for type
*anthropic.ToolUseBlockParam: json: error calling MarshalJSON for type
json.RawMessage: unexpected end of JSON input
------------------------
node path: [node_1, ChatModel]
```

This is the **same error and root cause as #462** (reported for the *empty* `arg` case and addressed by the `if args == ""` guard), and the same family as #304. The empty case is fixed; the **invalid/truncated** case is not.

**Root cause** (current `main`, `components/model/claude/claude.go`, around lines 1126–1134):

```go
args := tc.Function.Arguments
if args == "" {
    args = "{}"
}
// Arguments are limited to object type ...
messageParams = append(messageParams,
    anthropic.NewToolUseBlock(tc.ID, json.RawMessage(args), tc.Function.Name))
```

The guard catches `args == ""` but not `args` that is non-empty yet not valid JSON (e.g. `{"file_path":"/tmp/x.json","content":"{\"items\":[` — a `write_file` call cut off mid-`content`). `json.RawMessage(args)` keeps the partial bytes; `MarshalJSON` rejects them.

**Why the streamed `Arguments` can be invalid**

With streaming, tool-call input arrives as `input_json_delta` chunks that `ConcatMessageStream` folds into `Function.Arguments`. If the tool_use stream is cut before the JSON object closes (large payload, connection blip), the concatenated `Arguments` is syntactically incomplete. Per Anthropic's fine-grained tool streaming docs, partial tool_use deltas may not be valid JSON — so the consumer must tolerate invalid `Arguments`. The empty-case guard already acknowledges this class of problem; it just doesn't cover the non-empty-invalid sub-case.

**To Reproduce**

1. Use the Claude ChatModel in streaming mode with a tool whose input can be large (e.g. a `write_file` with a big `content`).
2. Have the model emit a tool call whose argument JSON gets truncated mid-stream (or, deterministically, hand-build a `schema.Message` with one `ToolCall` whose `Function.Arguments` is `{"a":` and run another turn so `convToolCalls` re-assembles it).
3. The next `Generate`/`Stream` fails with the `MarshalJSON ... unexpected end of JSON input` error above, ending the run as a framework error.

**Expected behavior**

A tool call with invalid `Arguments` should not crash request re-assembly. The empty-case guard should be generalized to **any** non-valid-JSON `Arguments`, mirroring the existing `if args == ""` fix:

```go
args := tc.Function.Arguments
if args == "" || !json.Valid([]byte(args)) {
    args = "{}"
}
messageParams = append(messageParams,
    anthropic.NewToolUseBlock(tc.ID, json.RawMessage(args), tc.Function.Name))
```

This keeps existing behavior for empty/valid args and replaces only the invalid case with a valid empty object so `MarshalJSON` succeeds. The model then receives a normal tool result (instead of a framework crash) and can retry. `encoding/json` is already imported in this file.

**Environment**
- eino-ext `components/model/claude` (observed on v0.1.18; same code on `main` / v0.1.20)
- anthropics/anthropic-sdk-go (`ToolUseBlockParam.MarshalJSON` over `json.RawMessage`)
- Go (any), streaming path

**Additional context — how this affects real consumers**

We hit this in production in a Go agent harness built on Eino's DeepAgent: an agent asked to write a large JSON file in one `write_file` call had its tool_use stream truncated, and the *entire turn crashed* (`status: failed`) rather than recovering. It happened across multiple unrelated tasks (a generic large-tool-input problem, not specific to one tool).

As a downstream workaround we wrapped the public `ToolCallingChatModel`/`BaseModel` interface (same composition pattern as Eino's own examples — no fork) and sanitize `ToolCalls[].Function.Arguments` on the model boundary before the message proceeds in the graph: any `Arguments` that fails `json.Valid` is replaced with a valid placeholder, so re-assembly never crashes, and we hand the model a tool result explaining the call was discarded so it can rewrite the call smaller. That fully fixes it for us — but it is a per-consumer wrapper for what is really a one-line generalization in `convToolCalls`. Fixing it upstream would protect every Eino+Claude user against the same crash for free, exactly as the `if args == ""` guard already does for the empty case.

A PR follows. Happy to adjust per your preference.

gutopolewacz:
### What

`convSchemaMessage` (`components/model/claude/claude.go`) only collapsed an **empty** tool-call `Arguments` to `"{}"` before wrapping it in `json.RawMessage`. A **non-empty but invalid** `Arguments` — typically a streamed `tool_use` truncated mid-payload (the `input_json_delta` chunks are cut before the JSON object closes) — was passed through raw, so `ToolUseBlockParam.MarshalJSON` failed on the next request re-assembly:

```
[NodeRunError] create new streaming message fail: ... *anthropic.ToolUseBlockParam:
json: error calling MarshalJSON for type json.RawMessage: unexpected end of JSON input
```

This is the same crash as #462 (empty arg, already fixed by the `if args == ""` guard) and the same family as #304 — but for the invalid/truncated sub-case, which the guard didn't cover. Closes #894.

### How

Generalize the existing guard to **any** non-valid-JSON `Arguments` via `json.Valid`:

```go
if args == "" || !json.Valid([]byte(args)) {
    args = "{}"
}
```

Valid JSON passes through unchanged; empty/invalid both collapse to a valid empty object so marshaling succeeds and the run continues (the model gets a normal tool result and can retry) instead of crashing. `encoding/json` is already imported.

### Tests

Added `Test_convSchemaMessage_ToolCallArguments` covering: valid args pass through, empty → `{}`, truncated → `{}`, malformed → `{}`, and asserting the assembled `MessageParam` marshals without error.

Note: I couldn't run the package test binary locally — `mockey v1.2.13` fails to build on my toolchain (`go1.26 / darwin-arm64`: `relocation target runtime.duffcopy not defined`), which is unrelated to this change (it fails for any test in the package). The non-test package builds and `go vet` is clean; CI should exercise the new test on a compatible toolchain. The new test does not use mockey.

### Why it helps others

We hit this in production in a Go agent harness on Eino's DeepAgent: an agent asked to write a large JSON file in one `write_file` had its tool_use stream truncate, crashing the whole turn. We worked around it downstream by wrapping the public model interface and sanitizing `Arguments` on the model boundary — but that's a per-consumer patch for what is a one-line generalization here. Fixing it upstream protects every Eino+Claude streaming user against the same crash, exactly as the empty-arg guard already does.

### DIFF
--- a/components/model/claude/claude.go
+++ b/components/model/claude/claude.go
@@ -1196,7 +1196,10 @@ func convSchemaMessage(message *schema.Message) (mp anthropic.MessageParam, err 
 		tc := message.ToolCalls[i]
 
 		args := tc.Function.Arguments
-		if args == "" {
+		// Empty or invalid (e.g. truncated streamed) Arguments would make
+		// ToolUseBlockParam.MarshalJSON fail on re-assembly; collapse both to a
+		// valid empty object (extends the empty-arg fix from #462).
+		if args == "" || !json.Valid([]byte(args)) {
 			args = "{}"
 		}
 		// Arguments are limited to object type.
--- a/components/model/claude/claude_test.go
+++ b/components/model/claude/claude_test.go
@@ -1269,3 +1269,46 @@ func TestVertexServiceAccountJSON(t *testing.T) {
 		})
 	})
 }
+
+func Test_convSchemaMessage_ToolCallArguments(t *testing.T) {
+	// A tool call whose Function.Arguments is empty or invalid JSON (e.g. a
+	// streamed tool_use truncated mid-payload) must not crash request
+	// re-assembly: convSchemaMessage collapses it to a valid empty object so
+	// ToolUseBlockParam.MarshalJSON succeeds. Valid JSON passes through unchanged.
+	cases := []struct {
+		name string
+		args string
+		want string
+	}{
+		{name: "valid passes through", args: `{"city":"Paris"}`, want: `{"city":"Paris"}`},
+		{name: "empty becomes empty object", args: "", want: "{}"},
+		{name: "truncated becomes empty object", args: `{"file_path":"/tmp/x.json","content":"{\"items\":[`, want: "{}"},
+		{name: "malformed becomes empty object", args: `{"a":}`, want: "{}"},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			msg := &schema.Message{
+				Role: schema.Assistant,
+				ToolCalls: []schema.ToolCall{{
+					ID: "call-1",
+					Function: schema.FunctionCall{
+						Name:      "write_file",
+						Arguments: tc.args,
+					},
+				}},
+			}
+
+			mp, err := convSchemaMessage(msg)
+			assert.NoError(t, err)
+			assert.Len(t, mp.Content, 1)
+
+			tu := mp.Content[0].OfToolUse
+			assert.NotNil(t, tu)
+			assert.Equal(t, tc.want, string(tu.Input.(json.RawMessage)))
+
+			// The whole message must marshal without "unexpected end of JSON input".
+			_, err = mp.MarshalJSON()
+			assert.NoError(t, err)
+		})
+	}
+}


gutopolewacz:
Rebased onto latest main and resolved the conflict in `claude_test.go` (both `TestVertexServiceAccountJSON` and the new `Test_convSchemaMessage_ToolCallArguments` now coexist). The fix in `claude.go` is unchanged. CI was green before the rebase — would appreciate a review when someone has a moment. Thanks!

PR #909, opened 2026-07-06
feat(callbacks/langfuse): expose max event size option

jesse-engineer:
## Summary
- Add `WithMaxEventSizeBytes` to `libs/acl/langfuse` so callers can configure the per-event truncation threshold.
- Expose `MaxEventSizeBytes` on `callbacks/langfuse.Config` and pass it through to the ACL client.
- Document the new option and cover custom/default truncation behavior with tests.

## Notes
- `callbacks/langfuse` uses a local replace for `libs/acl/langfuse` so this cross-module PR can compile against the sibling ACL change before a new ACL tag is published.

## Test plan
- `go test ./...` in `libs/acl/langfuse`
- `go test ./...` in `callbacks/langfuse`

### DIFF
--- a/callbacks/langfuse/README.md
+++ b/callbacks/langfuse/README.md
@@ -89,6 +89,10 @@ type Config struct {
     // MaxTaskQueueSize is the max number of events to buffer (Optional)
     // Default: 100
     MaxTaskQueueSize int
+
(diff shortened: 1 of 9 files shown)


jesse-engineer:
Closing this combined PR in favor of a split flow. The ACL-only change is now in #910; the callbacks/langfuse Config passthrough should follow after the ACL module change is merged/released.

PR #914, opened 2026-07-09
feat(model): support mid-conversation system messages

mrh997:
## Summary

Adds support for system messages appearing at any position in the conversation, not just at the beginning. This enables richer prompt engineering patterns where system instructions can be injected between user/assistant turns.

## API Changes

None. No exported types, functions, or fields were added, removed, or renamed.

## Key Changes

1. **Claude / AgenticClaude**: Leading system messages still go to the top-level `system` instruction field. Mid-conversation system messages are kept inline using `MessageParamRoleSystem` (SDK upgraded to v1.56.0). Consecutive inline system messages are merged into a single message with multiple content blocks.
2. **Gemini / AgenticGemini**: Leading system messages still populate `SystemInstruction`. Mid-conversation system messages are passed with role `"system"` directly in the content array. The `toGeminiRole` default case now returns an error instead of silently falling back to `"user"`.
3. **Qianfan**: The `toQianfanRole` function now passes `"system"` through as its own role instead of mapping it to `"user"`.

### DIFF
--- a/components/model/agenticclaude/convertor.go
+++ b/components/model/agenticclaude/convertor.go
@@ -26,27 +26,40 @@ import (
 	"github.com/cloudwego/eino/schema/claude"
 )
 
-func toAnthropicMessages(input []*schema.AgenticMessage) (systemBlocks []anthropic.TextBlockParam, msgParams []anthropic.MessageParam, err error) {
+func toAnthropicMessages(input []*schema.AgenticMessage) (sysInstruction []anthropic.TextBlockParam, msgParams []anthropic.MessageParam, err error) {
 	if len(input) == 0 {
 		return nil, nil, fmt.Errorf("input is empty")
 	}
 
-	systemDone := false
+	firstNonSystem := 0
+	for i, msg := range input {
+		if msg.Role != schema.AgenticRoleTypeSystem {
+			firstNonSystem = i
+			break
+		}
+		if i == len(input)-1 {
+			return nil, nil, fmt.Errorf("input contains only system messages")
+		}
+	}
 
-	for _, msg := range input {
+	for _, msg := range input[:firstNonSystem] {
+		blocks, err := toSystemBlocks(msg)
+		if err != nil {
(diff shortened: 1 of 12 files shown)


components/model/claude/claude.go
```diff
@@ -737,6 +734,22 @@ func mergeAdjacentToolResults(msgParams []anthropic.MessageParam) []anthropic.Me
 	return result
 }
 
+func mergeAdjacentSystemMessages(msgParams []anthropic.MessageParam) []anthropic.MessageParam {
```
shentongmartin, line 740: nit: `mergeAdjacentSystemMessages` 缺少 doc comment，旁边的 `mergeAdjacentToolResults` 有说明。建议补上一句，保持风格一致。

例如：
```go
// mergeAdjacentSystemMessages merges adjacent system-role messages into one
// with multiple content blocks, as required by the Anthropic API.
func mergeAdjacentSystemMessages(msgParams []anthropic.MessageParam) []anthropic.MessageParam {
```

components/model/agenticclaude/convertor.go
```diff
@@ -90,6 +103,22 @@ func mergeAdjacentToolResults(msgParams []anthropic.MessageParam) []anthropic.Me
 	return result
 }
 
+func mergeAdjacentSystemMessages(msgParams []anthropic.MessageParam) []anthropic.MessageParam {
```
shentongmartin, line 109: nit: 同 claude 包，`mergeAdjacentSystemMessages` 缺 doc comment。和 `mergeAdjacentToolResults` 保持一致，加一句说明。

components/model/claude/claude.go
mrh997: done

components/model/agenticclaude/convertor.go
mrh997: done

shentongmartin (approved):

PR #928, opened 2026-07-22
feat(claude): pass message.Extra through to callback output

sin-bufan:
## Summary

Forward `schema.Message.Extra` to `model.CallbackOutput.Extra` in `getCallbackOutput` so that provider-specific metadata stored in the message Extra map is available to callback handlers.

## Motivation

The `model.CallbackOutput` struct already has an `Extra map[string]any` field (added in eino v0.9.x), but the Claude adapter's `getCallbackOutput` does not populate it from the source `schema.Message.Extra`. This means any provider-specific metadata that a component stores in `message.Extra` (e.g. cache token breakdowns, model-specific diagnostics) is silently dropped before reaching callback handlers like Langfuse tracing.

## Changes

- **`components/model/claude/claude.go`**: Added 3 lines in `getCallbackOutput` to copy `output.Extra` into `result.Extra` when non-empty.

## Impact

- Zero breaking changes — the `Extra` field already exists on `CallbackOutput` but was unused by this adapter.
- Callback handlers that inspect `CallbackOutput.Extra` will now receive any metadata the Claude adapter stores in `schema.Message.Extra`.

## Test Plan

- Verified the change compiles successfully.
- Existing tests pass without modification.


sin-bufan:
Closing in favor of #929, which now includes this change (Extra pass-through in `getCallbackOutput`) alongside the cache token population in `convOutputMessage` and `AutoCacheControl` config field — all part of the same feature.

PR #932, opened 2026-07-23
fix(local): add IsDir checks to Read/Write/Edit to prevent directory-as-file errors

luchun19921229:
## Problem

When a user passes a directory path to `Local.Read()`, `Local.Write()`, or `Local.Edit()`, the methods attempt to use the path as a file — opening, reading, or writing it — which produces confusing low-level errors (e.g. `is a directory` from the OS) or silent failures instead of a clear explanation.

## Changes

Added a guard check in each of the three methods that returns a clear error message when the target path is a directory rather than a file:

### `Read()`
- Calls `file.Stat()` before `info.Size() == 0` check
- Returns `"path is a directory, not a file: <path>"`

### `Write()`
- Calls `os.Stat(path)` before `MkdirAll`/`OpenFile`
- Returns `"path is a directory, not a file: <path>"`

### `Edit()`
- Calls `os.Stat(path)` before `os.ReadFile(path)`
- Returns `"path is a directory, not a file: <path>"`

## Testing

- [ ] Run existing test suite to confirm no regressions
- [ ] Verify each method returns the directory error when given a directory path

### DIFF
--- a/adk/backend/local/local.go
+++ b/adk/backend/local/local.go
@@ -322,6 +322,9 @@ func (s *Local) Read(ctx context.Context, req *filesystem.ReadRequest) (*filesys
 	if err != nil {
 		return nil, fmt.Errorf("failed to stat file: %w", err)
 	}
+	if info.IsDir() {
+		return nil, fmt.Errorf("path is a directory, not a file: %s", path)
+	}
 	if info.Size() == 0 {
 		return &filesystem.FileContent{}, nil
 	}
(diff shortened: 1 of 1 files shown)


luchun19921229:
@cloudwego/eino-maintainers 请 review 这个 PR。CI 全绿，改动很简单：在 `adk/backend/local/local.go` 的 `Read`/`Write`/`Edit` 三个方法里加了 `IsDir()` 守卫，防止目录被误当成文件操作，给出清晰错误信息。

PR #936, opened 2026-07-28
feat(callbacks/tls): add opt-in TLS observability callback

### LINKED ISSUE #935: feat(callbacks/tls): add opt-in TLS observability callback
## Problem

Eino currently offers generic callback integrations, but applications that use Volcengine TLS need an opt-in, first-class path to export Eino callback spans to TLS while preserving the existing callback and OpenTelemetry integrations.

## Proposed design

Add a standalone `callbacks/tls` Go module that:

- builds Eino callback events into OpenTelemetry spans and maps them to TLS trace logs;
- keeps TLS export disabled by default, so existing users are unaffected;
- accepts credentials and endpoint configuration only through `TLSConfig` or environment variables, never hard-coded values;
- exposes an explicit exporter lifecycle (flush/shutdown) and leaves the generic OTLP exporter unchanged;
- models `agent.turn`, `llm.request`, and `tool.call` spans so traces remain queryable across text, multimodal, streaming, and tool-call flows.

Streaming output is coalesced before it is recorded, rather than persisting every chunk as a separate payload. Token usage, model/provider identity, tool arguments/results, errors, and duration are attached when available.

## Scope

This proposal also adds TLS connection options to `libs/acl/opentelemetry` and enriches the Ark model adapter with callback usage metadata. The implementation is intentionally additive and does not change Eino public behavior when TLS exporting is not configured.

## Validation

- unit tests for callback/span mapping, exporter configuration, streaming aggregation, tool calls, and error paths;
- integration coverage guarded by `TLS_EXPORTER_E2E=1`, so ordinary upstream CI does not require a TLS account;
- race-enabled tests for the affected `callbacks/tls`, `libs/acl/opentelemetry`, and Ark model modules.

## Alternatives

Users can assemble a custom callback handler or export generic OTLP spans. A dedicated module provides stable TLS resource/schema conventions and reduces repeated application-side integration work.

runninggo:
#### What type of PR is this?

feat: A new opt-in callback integration.

#### Check the PR title

- [x] This PR title matches the format: `<type>(optional scope): <description>`
- [x] The description is user-oriented and clear.
- [x] The new module includes English and Chinese usage/configuration READMEs.

#### More detailed description

en:

Adds `callbacks/tls`, an opt-in Eino callback module that exports callback spans to Volcengine TLS. It keeps existing callback and generic OTLP behavior unchanged unless TLS export is explicitly configured.

The module records `agent.turn`, `llm.request`, and `tool.call` spans, aggregates streaming chunks before exporting, and attaches token usage, provider/model metadata, tool arguments/results, error state, and duration when available.

The PR also adds TLS transport options in `libs/acl/opentelemetry` and Ark callback usage metadata needed for complete trace attribution.

The module-local `genproto` replacement was removed so downstream consumers do not rely on non-transitive Go module configuration.

zh(optional):

新增可选的 `callbacks/tls` 回调模块，将 Eino 回调 Span 导出到 Volcengine TLS；未显式配置时不会改变现有 callback 或通用 OTLP 行为。流式输出会在导出前聚合，并保留模型、Token、工具调用、错误和耗时信息。

#### Which issue(s) this PR fixes

Fixes #935

#### Validation

- `go test -mod=readonly -race -gcflags='all=-l -N' ./...` in `callbacks/tls`
- `go test -mod=readonly -race -gcflags='all=-l -N' ./...` in `libs/acl/opentelemetry`
- `go test -mod=readonly -race -gcflags='all=-l -N' ./...` in `components/model/ark`
- Apache 2.0 license headers, `gofmt`, and `git diff --check` verified.

Upstream GitHub Actions will run after this PR is opened.


callbacks/tls/tls_handler.go
```diff
@@ -0,0 +1,1494 @@
...
+	}
+	return result
+}
+
+func decodeJSONValue(raw string) any {
+	if raw == "" {
+		return ""
+	}
+	var value any
+	if json.Unmarshal([]byte(raw), &value) == nil {
```
shentongmartin: This `json.Unmarshal` into `any` converts JSON numbers to `float64`, so the later marshal in the Lens payload can round legitimate tool arguments. For example, `{"record_id":9007199254740993}` becomes `{"record_id":9007199254740992}`.

Please preserve the raw JSON (for example, `json.RawMessage`) or decode with `json.Decoder.UseNumber`, and add a regression test for an integer above `2^53`.

callbacks/tls/option.go
```diff
@@ -0,0 +1,61 @@
...
+	return func(o *options) {
+		if o.concatFuncs == nil {
+			o.concatFuncs = make(map[reflect.Type]any)
+		}
+
+		o.concatFuncs[reflect.TypeOf((*T)(nil)).Elem()] = fn
+	}
+}
+
+func WithAggrMessageOutput(enable bool) Option {
```
shentongmartin, line 60: `WithAggrMessageOutput(false)` currently has no observable effect. The value is stored in `defaultDataParser`, but `ParseStreamOutput` always calls `tryConcatOutputChunks` and exports the aggregated message.

Please branch the stream-output behavior on this option (and cover both settings), or remove the public option if aggregation must remain unconditional.

libs/acl/opentelemetry/option.go
```diff
@@ -156,6 +157,13 @@ func WithInsecure() Option {
 	})
 }
 
+// WithTLSInsecure disables client transport security for the exporter's gRPC with tls
```
shentongmartin: This option name and doc say it disables transport security, but the implementation passes `credentials.NewClientTLSFromCert(nil, "")`, which enables normal certificate-validating TLS and is effectively the OTLP client's default secure path.

Please remove or rename this no-op API to describe its real behavior. If an insecure transport is actually required, use the existing `WithInsecure` option explicitly rather than publishing a misleading TLS-specific alternative.

shentongmartin (approved):

PR #942, opened 2026-08-06
fix(ark): expose Responses API model name on messages

### LINKED ISSUE #941: ark: Responses API messages do not expose model name via ark.GetModelName
### Component

`components/model/ark`

### Problem

When using the Ark ChatModel with the Responses API enabled, the returned `schema.Message` does not expose the resolved model name via `ark.GetModelName(message)`.

The Chat Completions path does expose it, so application code that records per-call model usage has to handle the two APIs differently.

Observed behavior:

- Chat Completions API: `ark.GetModelName(message)` returns a value such as `doubao-seed-2-0-lite`.
- Responses API: `ark.GetModelName(message)` returns an empty string.
- Responses API callback output can contain `CallbackOutput.Extra["model_name"]` after #627, so the model name is available in callback metadata, but not on the returned message itself.

### Why this matters

For systems that aggregate token usage or model attribution from `schema.Message` metadata, switching Ark to Responses API silently loses the model identity. This makes Responses API less compatible with the Chat Completions adapter behavior.

The endpoint id is not a sufficient replacement because endpoint id and resolved model name have different semantics. Users may know only the endpoint id they configured, while billing/observability often needs the actual resolved model name.

### Expected behavior

`ark.GetModelName(message)` should return the resolved model name for Responses API responses as well, matching the Chat Completions path when the provider response contains model information.

If the intended design is that Responses API only exposes model name through callback extra, it would be helpful to document this difference explicitly.

### Current workaround

Read model name from callback output first, for example:

```go
modelName := ark.GetModelName(callbackOutput.Message)
if modelName == "" && callbackOutput.Extra != nil {
    if v, ok := callbackOutput.Extra["model_name"].(string); ok {
        modelName = v
    }
}
```

This works for callback-based usage after #627, but code paths that only receive the final `schema.Message` still cannot recover the model name consistently.

jesseedcp:
#### What type of PR is this?

fix

#### Check the PR title.

- [x] This PR title matches the format: `<type>(optional scope): <description>`
- [x] The description is user-oriented and clear.
- [x] No user documentation update is required; this makes the Responses API consistent with the existing Chat Completions message metadata.

`fix(ark): 在 Responses API 消息中暴露模型名称`

en:

The Ark Chat Completions path exposes the resolved model name on `schema.Message` through `ark.GetModelName`, while the Responses API currently reports it only through callback extra.

This change reuses the existing message-extra mechanism to preserve `ResponseObject.Model` on non-streaming output messages and streaming response metadata chunks. As a result, callers can retrieve the model name from both non-streaming messages and concatenated streaming messages through `ark.GetModelName`.

No public API or dependency changes are introduced.

Tests:

- `go test -gcflags="all=-N -l" ./...`
- `go vet ./...`
- `go test -race -gcflags="all=-N -l" . -run '^TestResponsesAPIChatModelPreservesModelName$' -count=1`

Fixes #941

#### (optional) The PR that updates user documentation:

N/A

### DIFF
--- a/components/model/ark/responses_api.go
+++ b/components/model/ark/responses_api.go
@@ -998,6 +998,7 @@ func (cm *ResponsesAPIChatModel) toOutputMessage(resp *responses.ResponseObject,
 	}
 	setContextID(msg, resp.Id)
 	setResponseID(msg, resp.Id)
+	setModelName(msg, resp.Model)
 
 	if resp.ServiceTier != nil {
 		setServiceTier(msg, resp.ServiceTier.String())
@@ -1257,6 +1258,7 @@ func (cm *ResponsesAPIChatModel) setStreamChunkDefaultExtra(msg *schema.Message,
 	}
 	setContextID(msg, object.Id)
 	setResponseID(msg, object.Id)
+	setModelName(msg, object.Model)
 	if object.ServiceTier != nil {
 		setServiceTier(msg, object.ServiceTier.String())
 	}
--- a/components/model/ark/responses_api_test.go
+++ b/components/model/ark/responses_api_test.go
@@ -75,6 +75,61 @@ func TestResponsesAPIChatModelGenerate(t *testing.T) {
 	})
 }
 
+func TestResponsesAPIChatModelPreservesModelName(t *testing.T) {
+	const modelName = "doubao-seed-2-0-lite"
+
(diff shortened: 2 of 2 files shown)


BrahmaMantra:
@hi-pender Sorry to bother you. Could you take a look at this PR and see if it makes sense? PTAL~

PR #929, opened 2026-07-22
feat(claude): expose cache_creation_input_tokens in message.Extra and add Config.AutoCacheControl

sin-bufan:
## Summary

Expose Anthropic's `cache_creation_input_tokens` usage metric in the Claude adapter, following the `arkbot` / `ark` / `gemini` pattern of using `message.Extra` for provider-specific data. Also add a `Config.AutoCacheControl` field for injecting automatic cache breakpoints at construction time.

### cache_creation_input_tokens

Anthropic bills prompt cache reads (~10% of base rate) and writes (~125% of base rate) at very different prices. The read side is already exposed through the standard `PromptTokenDetails.CachedTokens` path. The creation side (`cache_creation_input_tokens`) has no standard schema equivalent, so it is surfaced in `message.Extra`:

- **Key**: `_eino_claude_cache_creation_input_tokens` (`int`)
- **Set in**: `convOutputMessage` (Generate) and `convStreamEvent` (message_start, message_delta)
- **Read via**: `claude.GetCacheCreationInputTokens(msg *schema.Message) (int, bool)`
- **Streaming**: Anthropic reports once per request on `message_start`. `int` has built-in `useLast` concat, so `ConcatMessages` correctly preserves the value without accumulation.
- **Zero activity**: When `CacheCreationInputTokens == 0`, no Extra entry is written.

### Config.AutoCacheControl

When building a `ChatModel` once and handing it to ADK Runner, there is no per-request injection point for `model.Option`. `Config.AutoCacheControl` allows enabling automatic cache breakpoints at construction time, passed through to the SDK via `model.GetImplSpecificOptions` in `genParamsAndOptions`.

### Design decisions

- **Only `cache_creation` in Extra** — `cache_read` is NOT duplicated because it's already in the standard `PromptTokenDetails.CachedTokens` path. This avoids the "two sources of truth" problem.
- **`int` type, not a struct** — Simpler than a typed struct; no need to register concat func or schema name since `int` has built-in support.
- **`getCallbackOutput` unchanged** — No Extra is set on `CallbackOutput`. Callback handlers access cache data via `CallbackOutput.Message.Extra`, consistent with how arkbot exposes `BotUsage`.
- **No blanket `output.Extra → CallbackOutput.Extra`** — Avoids leaking message-level implementation state (thinking text, signatures, tool_search_events) into callback metadata.

## Changes

| File | Change |
|------|--------|
[...]

PR #978, opened 2026-08-28
feat(agenticclaude): expose cache_creation_input_tokens in message.Extra

sin-bufan:
## Summary

Port `cache_creation_input_tokens` from the legacy claude adapter to agenticclaude, following the pattern established in #929.

Anthropic bills prompt cache reads (~10% of base rate) and writes (~125% of base rate) at very different prices. The read side is already exposed through the standard `PromptTokenDetails.CachedTokens` path. The creation side (`cache_creation_input_tokens`) has no standard schema equivalent, so it is surfaced in `message.Extra`:

- **Key**: `_eino_claude_cache_creation_input_tokens` (same key as the legacy claude adapter)
- **Set in**: `toAgenticMessage` (Generate) and `toMessageStreamingChunk` (message_delta)
- **Read via**: `agenticclaude.GetCacheCreationInputTokens(msg *schema.AgenticMessage) (int, bool)`
- **Streaming**: Anthropic reports once per request on message_start or message_delta. `int` has built-in `useLast` concat, so `ConcatAgenticMessages` correctly preserves the value without accumulation.
- **Zero activity**: When `CacheCreationInputTokens == 0`, no Extra entry is written.

### Design decisions

- **Same Extra key as legacy adapter** — Downstream consumers (e.g. Langfuse cost tracking) do not need to distinguish between legacy and agentic adapters.
- **Only `cache_creation` in Extra** — `cache_read` is NOT duplicated because it's already in the standard `PromptTokenDetails.CachedTokens` path.
- **`int` type, not a struct** — Simpler than a typed struct; no need to register concat func since `int` has built-in support.

## Changes

| File | Change |
|------|--------|
| `consts.go` | Add exported `KeyOfCacheCreationInputTokens` constant |
| `convertor.go` | Set cache_creation in `toAgenticMessage` (Generate path) |
| `event_convertor.go` | Set cache_creation in `toMessageStreamingChunk` (message_delta) |
| `message_extra.go` | New: `GetCacheCreationInputTokens()` getter |
| `message_extra_test.go` | New: tests for getter, Generate path, streaming delta path |

## Test plan

- [x] `TestGetCacheCreationInputTokens` — nil message, nil Extra, key missing, key present
- [x] `TestCacheCreationInputTokens_SetInConvOutputMessage` — zero and nonzero paths
- [x] `TestCacheCreationInputTokens_SetInStreamDelta` — zero and nonzero paths
- [x] All existing tests pass (`go test ./...`)

## Related

- #929 — Same feature for the legacy claude adapter (already merged)

### DIFF
--- a/components/model/agenticclaude/consts.go
+++ b/components/model/agenticclaude/consts.go
@@ -88,8 +88,4 @@ const (
 
 const implType = "AgenticClaude"
 
-// KeyOfCacheCreationInputTokens is the Extra map key for Anthropic
-// cache_creation_input_tokens. It uses the same key as the legacy claude
-// adapter so downstream consumers (e.g. Langfuse cost tracking) do not need
-// to distinguish between the two adapters.
-const KeyOfCacheCreationInputTokens = "_eino_claude_cache_creation_input_tokens"
+const keyOfCacheCreationInputTokens = "_eino_claude_cache_creation_input_tokens"
--- a/components/model/agenticclaude/consts.go
+++ b/components/model/agenticclaude/consts.go
@@ -87,3 +87,9 @@ const (
 )
 
 const implType = "AgenticClaude"
+
+// KeyOfCacheCreationInputTokens is the Extra map key for Anthropic
+// cache_creation_input_tokens. It uses the same key as the legacy claude
+// adapter so downstream consumers (e.g. Langfuse cost tracking) do not need
+// to distinguish between the two adapters.
+const KeyOfCacheCreationInputTokens = "_eino_claude_cache_creation_input_tokens"
--- a/components/model/agenticclaude/message_extra.go
+++ b/components/model/agenticclaude/message_extra.go
@@ -1,0 +1,39 @@
+/*
+ * Copyright 2026 CloudWeGo Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package agenticclaude
+
+import "github.com/cloudwego/eino/schema"
+
+// GetCacheCreationInputTokens returns the Anthropic cache_creation_input_tokens
+// count from the message. This is the number of input tokens written into a new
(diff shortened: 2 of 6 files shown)


components/model/agenticclaude/consts.go
```diff
@@ -87,3 +87,9 @@ const (
 )
 
 const implType = "AgenticClaude"
+
+// KeyOfCacheCreationInputTokens is the Extra map key for Anthropic
+// cache_creation_input_tokens. It uses the same key as the legacy claude
+// adapter so downstream consumers (e.g. Langfuse cost tracking) do not need
+// to distinguish between the two adapters.
+const KeyOfCacheCreationInputTokens = "_eino_claude_cache_creation_input_tokens"
```
shentongmartin: nit: 这里建议先别把 raw Extra key 暴露成 public API。老的 `claude` adapter 里对应 key 是私有的，只通过 `GetCacheCreationInputTokens` 暴露读取能力；这样用户不会直接依赖 map key 这个实现细节，后续也更好调整。可以考虑改成 unexported `keyOfCacheCreationInputTokens`，保留 getter 作为唯一公开入口。

components/model/agenticclaude/message_extra.go
```diff
@@ -0,0 +1,39 @@
...
+// count from the message. This is the number of input tokens written into a new
+// prompt cache entry, billed at ~125% of the base input token rate.
+//
+// The cache-read side is available through the standard token usage path:
+//
+//	msg.ResponseMeta.TokenUsage.PromptTokenDetails.CachedTokens
+//
+// When streaming, Anthropic reports this once per request (on message_start or
+// message_delta). schema.ConcatAgenticMessages merges Extra maps with a
+// last-value-wins policy for int, so the final concatenated message carries
```
shentongmartin, line 31: nit: 这里文档依赖了 `schema.ConcatAgenticMessages` 对 `int` Extra 的 last-value-wins 行为，这个判断本身是对的，但最好补一个回归测试直接锁住这个契约。比如构造带 cache creation Extra 的 stream chunk，再拼一个普通文本 chunk，最后 concat 后用 `GetCacheCreationInputTokens` 断言还能读到原值。

components/model/agenticclaude/consts.go
sin-bufan: 已按建议调整：将 KeyOfCacheCreationInputTokens 改为私有的 keyOfCacheCreationInputTokens，对外仅保留 GetCacheCreationInputTokens 作为读取入口。

components/model/agenticclaude/message_extra.go
sin-bufan: 已补充回归测试 TestCacheCreationInputTokens_PreservedAfterConcat：构造包含 cache creation token 的 stream chunk，与普通文本 chunk 拼接后，断言仍能通过 GetCacheCreationInputTokens 读取原值。

shentongmartin (approved):

PR #983, opened 2026-09-02
feat(agenticopenai): expose prompt cache write usage

sin-bufan:
#### What type of PR is this?

feat

#### Check the PR title.

- [x] This PR title matches the required format.
- [x] The description is user-oriented and clear.
- [x] No separate user documentation update is required; this exposes response billing metadata to downstream integrations.

#### (Optional) Translate the PR title into Chinese.

feat(agenticopenai): 暴露提示词缓存写入用量

#### (Optional) More detailed description for this PR(en: English/zh: Chinese).

en:

OpenAI Responses for GPT-5.6 and later reports usage.input_tokens_details.cache_write_tokens separately from cached_tokens because cache writes and reads have different billing rates. agenticopenai currently preserves only cache reads, so downstream observability integrations cannot calculate prompt-cache cost accurately.

This change follows the message.Extra pattern established by #978:

- Exposes positive cache_write_tokens as _eino_openai_cache_write_tokens.
- Adds GetCacheWriteTokens for downstream consumers.
- Covers both Generate and response lifecycle streaming events.
- Keeps cache writes separate from PromptTokenDetails.CachedTokens.
- Reads the field from the OpenAI SDK preserved JSON ExtraFields until the SDK adds a typed property.
- Omits absent, zero, negative, malformed, fractional, and overflowing values.
- Verifies the value survives schema.ConcatAgenticMessages.

Tests:

- go test -gcflags=all=-N\\ -l ./...

zh(optional):

OpenAI Responses 的 cache_write_tokens 与 cached_tokens 使用不同计费价格。本改动参考 #978，通过 AgenticMessage.Extra 同时在 Generate 和流式响应中透传缓存写入 token，供 Langfuse 等下游准确计算成本。

#### (Optional) Which issue(s) this PR fixes:

Related to #978.

PR #1009, opened 2026-09-22
fix(agenticopenai): copy streamed messages before marking them auto-cached

sin-bufan:
#### What type of PR is this?

fix

#### Check the PR title.

fix(agenticopenai): 流式响应打 auto-cache 标记前先复制消息

en:
`ResponsesModel.Stream` calls `OnEndWithStreamOutput` before `setAutoCached`. `StreamReader.Copy` then shares the same `*AgenticMessage` with callbacks and the caller. `setAutoCached` writes `Message.Extra` on that shared pointer while callbacks read it, which is a fatal `concurrent map read and map write` (not a recoverable panic).

This copies the message and its `Extra` map before writing `_eino_ext_agenticopenai_auto_cached`. Callbacks keep the published map. The caller still receives the auto-cache mark, so prompt-cache continuation is unchanged. `Generate` is unchanged: it writes the flag before `OnEnd`, on the same goroutine, with no stream copy.

zh(optional):
`Stream` 在 `OnEndWithStreamOutput` 之后对已经分发给回调的消息原地写 `Extra`。`Copy` 只共享指针，并发读写这张 map 会让进程直接 fatal。现在先浅拷贝消息并复制 `Extra`，只给调用方那份打标。

This is independent of #983. That change assigns a newly allocated `Extra` map before `Send`, so it does not mutate a map that callbacks are already reading. The two PRs touch different lines and can merge in either order.


N3kox (dismissed):

N3kox:
@sin-bufan 

Overall this looks correct and nicely minimal — the race is real and the fix aligns precisely with how `StreamReader.Copy` shares the same pointer across branches. A few points worth noting (none blocking):

**1. Behavioral difference between `Stream` and `Generate` regarding callback visibility.**
After this fix, the auto-cache mark lives only on the caller's clone, so callbacks no longer observe `_eino_ext_agenticopenai_auto_cached` on the published message. This introduces an asymmetry with `Generate`, where `setAutoCached` runs before `OnEnd` on the same goroutine, so callbacks *do* see the mark. Since this flag is an internal detail meant for prompt-cache continuation, I'd argue the new behavior is actually cleaner — callbacks shouldn't depend on it. That said, the previous behavior exposed it non-deterministically and now it's deterministically hidden, so if any internal callback consumes this key it's worth being aware of. It might be helpful to call out this Generate/Stream difference explicitly in the PR description.

**2. The clone happens per-chunk.**
The caller's convert runs `markAutoCachedForCaller` for every non-nil chunk, so in the streaming path this is a per-chunk clone rather than a one-time copy. The cost is just a shallow struct copy plus a small `Extra` map clone, so it's negligible — and the original code already marked per-chunk, so it's the same order of magnitude. Just flagging it for visibility, no change needed.


Thanks for the clear write-up on the root cause — LGTM.

mrh997:
Thank you for reporting this and for putting together the PR. The issue is now addressed in #1011, which applies the fix comprehensively across multiple model implementations. We’ll close this PR in favor of that broader change. Thanks again for your feedback and contribution!

